PR #14, opened 2023-11-09
refactor(template): remove prettier config

elbalexandre:
# What does this PR do?

- remove prettier config from template


elbalexandre:
We don't need the prettier config in the template as its already in the root folder

proquig (approved):

PR #17, opened 2023-11-09
feat(Microsoft): implement Microsoft TPA

lweinhard:


apps/microsoft/app/microsoft/tpa.ts
```diff
@@ -0,0 +1,203 @@
...
+  thirdPartyAppsObjects: ThirdPartyAppsObjectsUpsertInput;
+  permissionGrantsObjects: PermissionGrantInsertInput[];
+} => {
+  const permissionGrantsObjects: PermissionGrantInsertInput[] = [];
+  const apps = [
+    // Will be used whenever we decide to scan apps that require application permissions
+    // ...servicePrincipalsWithScopes.map((sp) =>
+    //   formatServicePrincipalToThirdPartyAppObject(sp)
+    // ),
+    ...permissionGrants.map((pg) => {
```
elbalexandre, line 51: Why spreading the array ?

apps/microsoft/app/microsoft/tpa.ts
```diff
@@ -0,0 +1,203 @@
...
+      permissionGrantsObjects.push({
+        tenantId,
+        userId: pg.principalId,
+        appId: pg.clientId,
+        grantId: pg.id,
+      });
+      return formatPermissionGrantToThirdPartyAppObject(
+        pg,
+        allServicePrincipals.filter((sp) => sp.id === pg.clientId)[0]
```
elbalexandre, line 60: This array entry is uncheck. My bad, I did not push the ts rule in the config.

apps/microsoft/app/microsoft/tpa.ts
```diff
@@ -0,0 +1,203 @@
...
+    accessToken,
+    tenantId,
+    pageLink,
+  });
+  const allServicePrincipals = await getAllServicePrincipalsById({
+    accessToken,
+    tenantId,
+  });
+
+  // Will be used whenever we decide to scan apps that require application permissions
```
elbalexandre, line 133: This can be removed

apps/microsoft/app/schemas/syncJob.ts
```diff
@@ -0,0 +1,25 @@
...
+export const SyncJobTypeEnum = pgEnum('sync_job_type', SyncJobTypeEnumValues);
+
+export type SyncJobType = (typeof SyncJobTypeEnumValues)[number];
+
+export const syncJobs = pgTable('sync_job', {
+  id: uuid('id').primaryKey().defaultRandom(),
+  tenantId: text('tenant_id')
+    .notNull()
+    .references(() => organizations.tenantId),
+  status: SyncJobStatusEnum('status').default('scheduled').notNull(),
```
elbalexandre, line 17: This property is unused.

apps/microsoft/app/styles/globals.css
```diff
@@ -0,0 +1,78 @@
+@import 'tailwindcss/base';
```
elbalexandre, line 1: all the style can be removed for now

apps/microsoft/app/api/scan/third-party-apps/route.ts
```diff
@@ -0,0 +1,36 @@
+import { db } from '#src/lib/db';
+import { scanThirdPartyAppsByTenantId } from '#src/microsoft/tpa';
+import { organizations } from '#src/schemas/organization';
+import { syncJobs } from '#src/schemas/syncJob';
+import { eq, asc } from 'drizzle-orm';
+
+export const runtime = 'edge';
+
+export async function POST(_req: Request): Promise<Response> {
```
elbalexandre, line 9:
```suggestion
export async function POST(): Promise<Response> {
```

apps/microsoft/app/api/scan/users/route.ts
```diff
@@ -0,0 +1,33 @@
...
+    } else {
+      await db.insert(syncJobs).values({
+        tenantId: orgToSync[0].tenantId,
+        type: 'users',
+        paginationToken: result.pageLink,
+      });
+    }
+    return new Response(JSON.stringify(result), { status: 200 });
+  } catch (e) {
+    return new Response(e, { status: 500 });
```
elbalexandre, line 31: We could not handle Response errors for now, this could everywhere with a single middleware.

apps/microsoft/app/api/scan/users/route.ts
```diff
@@ -0,0 +1,33 @@
...
+        .set({ lastUserScan: new Date() })
+        .where(eq(organizations.id, orgToSync[0].id));
+    } else {
+      await db.insert(syncJobs).values({
+        tenantId: orgToSync[0].tenantId,
+        type: 'users',
+        paginationToken: result.pageLink,
+      });
+    }
+    return new Response(JSON.stringify(result), { status: 200 });
```
elbalexandre, line 29:
```suggestion
    return Response.json(result, { status: 200 });
```

This can be done everywhere. I am not sure, but the default could be `200`, so perhaps it's not even needed.

apps/microsoft/app/microsoft/auth.ts
```diff
@@ -0,0 +1,32 @@
...
+import { organizations } from '#src/schemas/organization';
+
+export const handleMicrosoftAuthCallback = async ({
+  tenantId,
+  isAdminConsentGiven,
+}: {
+  tenantId: string | null;
+  isAdminConsentGiven: boolean;
+}) => {
+  if (isAdminConsentGiven && tenantId) {
```
elbalexandre, line 13: An early return here could make the function less deep.

apps/microsoft/app/microsoft/auth.ts
```diff
@@ -0,0 +1,32 @@
...
+      !scopes.includes('DelegatedPermissionGrant.ReadWrite.All') ||
+      !scopes.includes('Application.ReadWrite.All') ||
+      !scopes.includes('User.Read.All')
+    ) {
+      throw new Error("Couldn't retrieve required scopes");
+    }
+    try {
+      await db.insert(organizations).values({ tenantId }).onConflictDoNothing();
+    } catch {
+      return 'You have already given admin consent. You may close this window now.';
```
elbalexandre, line 26: as there is `.onConflictDoNothing();`, if it's throw it should not be for that reason.

apps/microsoft/app/microsoft/tpa.ts
elbalexandre: let's use a `for .. of` that fill both the array, it will be clearer

apps/microsoft/next.config.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  reactStrictMode: true,
+  transpilePackages: ["ui"],
```
elbalexandre, line 3: There is no ui package for now
```suggestion
```

apps/microsoft/src/app/api/inngest/third-party-apps/schedule-third-party-apps-scans.ts
```diff
@@ -0,0 +1,24 @@
...
+  { cron: '0 0 * * *' },
+  async () => {
+    const orgs = await db
+      .select({
+        tenantId: organizations.tenantId,
+      })
+      .from(organizations);
+    await Promise.all(
+      orgs.map(({ tenantId }) =>
+        inngest.send({
```
elbalexandre, line 16: we should use `step.send` inside inngest function.
We can, as well, send an array of event instead of doing a `Promise.all`.

apps/microsoft/src/app/api/inngest/users/start-users-scan.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    });
+
+    expect(step.sendEvent).toBeCalledTimes(1);
+    expect(step.sendEvent).toBeCalledWith('scan-users', {
+      name: 'users/scan',
+      data: {
+        accessToken: 'token',
+        tenantId: mockedOrganization.tenantId,
+        organizationId: mockedOrganization.elbaOrganizationId,
+        syncStartedAt: new Date(event.ts).toISOString(),
```
elbalexandre, line 55: We've agreed on using `Date.now()`  instead of `event.ts`

apps/microsoft/src/common/microsoft.ts
```diff
@@ -0,0 +1,251 @@
...
+  const response = await fetch(
+    pageLink ??
+      `https://graph.microsoft.com/v1.0/${tenantId}/servicePrincipals?$top=${MAX_RESULTS_PER_PAGE}&$select=appDisplayName,description,id,homepage,info,verifiedPublisher,tags,appRoles&$expand=appRoleAssignments`,
+    {
+      headers: {
+        Authorization: `Bearer ${accessToken}`,
+      },
+    }
+  );
+
```
elbalexandre, line 145: On each of theses fetch calls, the status is not checked.
You can do it by doing this:

```
if (!response.ok) {
  throw new Error('Could not retrieve foo bar with ....')
}
```
Note that `fetch` is not throwing error when the status is not 20X. It's doing it, for example, when the server is not responding.

Because you can't really throw `FetchError`, you can create a custom error class `MicrosoftError` that have a `response` property so you are able to check the status and headers in middlewares.

apps/microsoft/src/common/microsoft.ts
```diff
@@ -0,0 +1,251 @@
...
+  return (await response.json()) as MicrosoftGraphAPIResponse<SafeMicrosoftGraphServicePrincipal>;
+};
+
+export const getAllServicePrincipalsById = async ({
```
elbalexandre, line 149: How many page we can expect here ? Should it be paginated using inngest ?

apps/microsoft/src/repositories/microsoft/auth.ts
```diff
@@ -0,0 +1,38 @@
...
+  isAdminConsentGiven,
+}: {
+  tenantId: string | null;
+  elbaOrganizationId: string | undefined;
+  isAdminConsentGiven: boolean;
+}) => {
+  if (!isAdminConsentGiven || !tenantId || !elbaOrganizationId) {
+    throw new Error('You must give admin consent to continue');
+  }
+  await timeout(10000);
```
elbalexandre, line 19: Why ?

apps/microsoft/src/repositories/microsoft/tpa.ts
```diff
@@ -0,0 +1,201 @@
...
+  thirdPartyAppsObjects: ThirdPartyAppsObjectsUpsertInput;
+  permissionGrantsObjects: PermissionGrantInsertInput[];
+} => {
+  const permissionGrantsObjects: PermissionGrantInsertInput[] = [];
+  const apps = [
+    // Will be used whenever we decide to scan apps that require application permissions
+    // ...servicePrincipalsWithScopes.map((sp) =>
+    //   formatServicePrincipalToThirdPartyAppObject(sp)
+    // ),
+    ...permissionGrants.value.map((pg) => {
```
elbalexandre, line 41: A `.map` that include a `push` followed by a `reduce`. That's looks complex. (why destructuring the array returned ?)

proposal:

```
const servicePrincipals = new Map<string, ServicePrincipal>(allServicePrincipals.map(service => [service.id, service]))
const apps = new Map<string, App>()
const permissionGrantsObjects = []
for (const permissionGrant of permissionGrants) {
  permissionGrantsObjects.push({
    tenantId,
    userId: pg.principalId,
    appId: pg.clientId,
    grantId: pg.id,
  });
  const  servicePrincipal = servicePrincipals.get(permissionGrant.clientId)
  if (!servicePrincipal) throw ...
  apps.set(servicePrincipal.id, formatPermissionGrantToThirdPartyAppObject(permissionGrant, servicePrincipal))
}

return {
  apps: Array.from(apps.values()),
  permissionGrantsObjects
}
```

Also why throws when a servicePrincipal is missing, can we just ignore it along the app it's related ?

apps/microsoft/src/schemas/permission-grant.ts
```diff
@@ -0,0 +1,22 @@
+import type { InferInsertModel, InferSelectModel } from 'drizzle-orm';
+import { pgTable, text, unique, uuid } from 'drizzle-orm/pg-core';
+import { organizations } from './organization';
+
+export const permissionGrants = pgTable(
+  'permission_grant',
+  {
+    id: uuid('id').notNull().primaryKey().defaultRandom(),
```
elbalexandre, line 8: Why not using the grant_id as primary key ?

apps/microsoft/src/schemas/organization.ts
```diff
@@ -0,0 +1,11 @@
+import type { InferInsertModel, InferSelectModel } from 'drizzle-orm';
+import { pgTable, text, uuid } from 'drizzle-orm/pg-core';
+
+export const organizations = pgTable('organization', {
+  id: uuid('id').notNull().primaryKey().defaultRandom(),
+  elbaOrganizationId: text('elba_organization_id').unique().notNull(),
```
elbalexandre, line 6: why not using organisation_id from elba as `id` (primary key) ?

apps/microsoft/src/common/microsoft.ts
```diff
@@ -0,0 +1,251 @@
...
+    {
+      headers: {
+        Authorization: `Bearer ${accessToken}`,
+      },
+    }
+  );
+  return (await response.json()) as MicrosoftGraphAPIResponse<SafeMicrosoftGraphPermissionGrant>;
+};
+
+export const getServicePrincipalAppRoleAssignedToById = async ({
```
elbalexandre, line 107: why this is not in `/repositories` folder ?

apps/microsoft/src/app/api/inngest/third-party-apps/scan-third-party-apps.ts
```diff
@@ -0,0 +1,67 @@
...
+  {
+    event: 'third-party-apps/scan',
+    rateLimit: {
+      key: 'event.data.tenantId',
+      limit: 1,
+      period: '24h',
+    },
+  },
+  async ({ event, step }) => {
+    const { tenantId, syncStartedAt, organizationId, cursor, accessToken } = event.data;
```
elbalexandre, line 23: IMO it's a bad idea to pass accessToken, if it's get refreshed or the function delayed we may be stuck with the old one.

apps/microsoft/src/app/api/inngest/third-party-apps/start-third-party-apps-scan.ts
```diff
@@ -0,0 +1,35 @@
...
+  },
+  {
+    event: 'third-party-apps/start',
+  },
+  async ({ event, step }) => {
+    const syncStartedAt = Date.now();
+    const { tenantId, isFirstScan } = event.data;
+
+    const [organization, token] = await step.run('initialize', () =>
+      Promise.all([getOrganizationByTenantId(tenantId), getTokenByTenantId(tenantId)])
```
elbalexandre, line 20: What do you think of using one event to run a scan. I did this implementation first, but it's just bring another event for few (no ?) benefits

apps/microsoft/src/schemas/organization.ts
elbalexandre: Up + I suggest that we name this table `organisation` to be consistent with elba API.

apps/microsoft/src/app/api/inngest/client.ts
```diff
@@ -0,0 +1,40 @@
...
+        accessToken: z.string(),
+        organizationId: z.string(),
+        syncStartedAt: z.string(),
+        cursor: z.string().optional(),
+        isFirstScan: z.boolean().default(false),
+      }),
+    },
+    'third-party-apps/start': {
+      data: z.object({
+        tenantId: z.string(),
```
elbalexandre, line 25: It could be PK of organisation here (`organisationId` as suggested) - it would be easier to debug.

apps/microsoft/src/app/api/webhook/sync-organization/service.ts
```diff
@@ -0,0 +1,29 @@
...
+import { organizations } from '@/schemas';
+import { db } from '@/lib/db';
+import { inngest } from '../../inngest/client';
+
+export const syncOrganization = async (organizationId: string) => {
+  const [organization] = await db
+    .select({
+      tenantId: organizations.tenantId,
+    })
+    .from(organizations)
```
elbalexandre, line 11: In drizzle doc they are using lowercase first-char word, but in reality this is not convenient when we are retrieving multiple items from a same table as var names are conflicting. I would suggest renaming it with an uppercase first char.

PR #31, opened 2023-11-21
zoom integration

baxsm:
## Description
Zoom auth / integration
## Checklist:


apps/zoom/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "next/core-web-vitals"
```
elbalexandre, line 2:
```suggestion
  "extends": "'custom/next'"
```

The project eslint configuration should be used here. If it's not working, you can refer to the `github` branch configuration. We might have to update the template.

apps/zoom/app/api/auth/getAccessToken/route.ts
```diff
@@ -0,0 +1,120 @@
...
+                ${userData.scope},
+                ${organization_id}
+            )
+        `;
+        return NextResponse.json("Created user information", { status: 200 });
+      }
+    } catch (error) {
+      return NextResponse.json(error, { status: 400 });
+    }
+  } catch (error: any) {
```
elbalexandre, line 117: We can just throw error in this file. A middleware will catch them and build a response from it.
If you still want to details error you can follow this pattern:

```
try {

} catch(error) {
  throw new Errror('Some detailed explanation...', { cause: error });
}
```

This comment applies to every response error in this PR.

apps/zoom/app/api/auth/refreshAccessToken/route.ts
```diff
@@ -0,0 +1,99 @@
...
+    });
+  }
+
+  try {
+    // getting user matching the following parameters
+    const getExistingUser = await sql`
+        SELECT * FROM zoom_credentials WHERE organization_id=${organization_id} AND refresh_token=${refresh_token};
+    `;
+    if (getExistingUser && getExistingUser.rows.length > 0) {
+      const base64EncodedBody = Buffer.from(
```
elbalexandre, line 61:
```suggestion
      const base64EncodedBearer = Buffer.from(
```
Let's called it by its name: `bearer`.

apps/zoom/app/api/auth/zoom/route.ts
```diff
@@ -0,0 +1,31 @@
...
+  const organization_id = req.nextUrl.searchParams.get("organization_id");
+
+  if (!organization_id || organization_id.length === 0) {
+    return NextResponse.json("organization_id not found in the query", {
+      status: 400,
+    });
+  }
+
+  // creating auth url for the user to redirect to
+  const getAuthUrl = () => {
```
elbalexandre, line 16: This can be moved in a `service.ts` that could be covered by some unit test.

apps/zoom/app/api/users/cron/run-sync-job/service.ts
```diff
@@ -0,0 +1,129 @@
...
+            const BATCH_SIZE = parseInt(process.env.USERS_SYNC_JOB_BATCH_SIZE || "5") || firstScheduledJob.rows[0].batchSize || 5;
+
+            const users = [];
+
+            for (let x = 0; x < batchUsers.rowCount; x += BATCH_SIZE) {
+                const batch = batchUsers.rows.slice(x, x + BATCH_SIZE);
+                for (const user of batch) {
+                    try {
+                        // get user information from zoom
+                        const response = await axios.get('https://api.zoom.us/v2/users/me', {
```
elbalexandre, line 41: I think there an issue here.
Only the Zoom admin will connect to the integration, not the other collaborators.

We need to use another endpoint to list every users of the organisation. If the rest API does not offer this possibility, you can check the zoom graphql API.

elbalexandre (commented):
I am aborting the review process, as the user fetching logic does not match our requirement. You can ask for another review, if you find a solution to the problem pointed out regarding zoom users retrieval.

Thanks for your contribution, overall the PR looks good.

PR #36, opened 2023-11-22
Notion - API

ApolloServices:


apps/notion-api/src/app/api/authentication/getAccessToken/route.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export async function POST (req: NextRequest) {
+        
+    const body : RequestBody = await req.json();
+    const { accessCode, organizationId } = body;
+
+    const encoded = Buffer.from(`${getNotionInfo().notionClientID}:${getNotionInfo().notionClientSecret}`).toString("base64");
+
+    const response = await fetch("https://api.notion.com/v1/oauth/token", {
```
athielba, line 26: I noticed that Notion has [SDK](https://github.com/makenotion/notion-sdk-js/blob/main/src/Client.ts#L78), can't we use it?

apps/notion-api/prisma/migrations/20231121154144_init/migration.sql
```diff
@@ -0,0 +1,22 @@
...
+
+    CONSTRAINT "Integration_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "Users_Sync_Jobs" (
+    "id" SERIAL NOT NULL,
+    "priority" INTEGER NOT NULL,
+    "organization_id" TEXT NOT NULL,
+    "pagination_token" TEXT NOT NULL,
```
athielba, line 18: `pagination_token` should be`nullable`

apps/notion-api/src/app/api/authentication/getAccessToken/route.ts
```diff
@@ -0,0 +1,80 @@
...
+                workspace_id: workspaceID,
+                workspace_name: workspaceName,
+                notionToken: accessToken,
+                },
+            });
+
+            await prisma.users_Sync_Jobs.create({
+                data: {
+                    priority: 1,
+                    pagination_token: '',
```
athielba, line 58: if `pagination_token` is `nullable` then no need to update `pagination_token`

apps/notion-api/src/app/api/users/cron/run-sync-user/service.ts
```diff
@@ -0,0 +1,129 @@
...
+    try {
+        const job = await prisma.users_Sync_Jobs.findFirst({
+            orderBy: [
+                { priority: 'asc' }, // Order by priority in ascending order
+                { sync_started_at: 'asc' }, // Order by timestamp in ascending order
+            ],
+        });
+
+        if (!job) {
+            return;
```
athielba, line 38: A proper message should be returned

apps/notion-api/src/app/api/users/cron/run-sync-user/service.ts
```diff
@@ -0,0 +1,129 @@
...
+        });
+
+        const data : UsersResponseType = await response.json();
+        if ('status' in data) {
+            return;
+        }
+
+        const users: UserData[] = [];
+        for (const user of data.results) {
+            const userData: UserData = {
```
athielba, line 73: User type can be `user` & `bot`, I believe we need to ignore the bot @n2ctech 

[Doc](https://developers.notion.com/reference/user)
Screenshot 2023-11-22 at 7 42 07 PM

apps/notion-api/src/app/api/users/cron/run-sync-user/service.ts
```diff
@@ -0,0 +1,129 @@
...
+        const nextCursor = data.next_cursor;
+
+        if (nextCursor) {
+            await prisma.users_Sync_Jobs.update({
+                where: {
+                    id: job.id,
+                },
+                data: {
+                    sync_started_at: new Date(),
+                    pagination_token: nextCursor ? nextCursor : '',
```
athielba, line 103: Condition is not required since it is in the `if`

apps/notion-api/src/app/api/users/cron/run-sync-user/service.ts
```diff
@@ -0,0 +1,129 @@
...
+                headers: {
+                accept: 'application/json',
+                'content-type': 'application/json',
+                },
+                data: { users, organisationId: integration.organization_id , sourceId},
+            };
+            await axios.request(options);
+        }
+    } catch (error) {
+
```
elbalexandre, line 125: Here is some pseudo code on how handle error in this case:
```suggestion
        if (error is NotionAPIError && error.retryAfter) {
          // update the job retry_after with the retrieved value
          return { message: 'Job delayed due to Notion rate limit', job };
        }
        // handle unauthorized access - the user might have disable the integration on notion
        if (error is NotionAPIError && error.response.status === 401) {
          // send a request to update the connection status on elba
          // delete the current job
          return { message: 'Job removed due to Notion unauthorized access', job }
        }
        // for the others (unexpected/unhandled) errors, we will increment the job retry count
        if (job.retry_count < process.env.MAX_USERS_SYNC_JOBS_RETRY_COUNT) {
          // increment job retry_count
          return { message: 'Job canceled due to retry limit', job }
        }
        // delete job
        throw error;
```

[Here is the documentation](https://elba-security.readme.io/reference/update-status) about /connection-status elba endpoint

apps/notion-api/src/app/api/authentication/getAccessToken/route.ts
```diff
@@ -0,0 +1,80 @@
...
+        body: JSON.stringify({
+        grant_type: "authorization_code",
+        code: accessCode,
+        redirect_uri: getNotionInfo().notionRedirectUrl
+        }),
+    });
+
+    const data : AccessTokenResponse = await response.json();
+
+    if (response.status === 200) {
```
elbalexandre, line 42: Prefer [early return](https://gomakethings.com/the-early-return-pattern-in-javascript/) here 
```suggestion
    if (!response.ok === 200) {
      throw new Error(`Could not retrieve access token from notion for organizationId=${organizationId}`)
    }
```

apps/notion-api/src/app/api/authentication/getAccessToken/route.ts
```diff
@@ -0,0 +1,80 @@
...
+            return NextResponse.json({
+                organization_id: organizationId,
+                workspace_id: workspaceID,
+                workspace_name: workspaceName,
+                notionToken: accessToken,
+            }, {
+                status: 200
+            });
+        } catch (error) {
+            return new NextResponse(null, { status: 500, statusText: 'Internal Server Error' });
```
elbalexandre, line 73: Just let throw the error here. We will handle error response in a middleware later.
```suggestion
```

apps/notion-api/src/app/api/users/cron/run-sync-user/route.ts
```diff
@@ -0,0 +1,7 @@
+import { NextResponse } from 'next/server';
+import { runUsersSyncJob } from './service';
+
+export async function GET() {
+    await runUsersSyncJob();
+    return new NextResponse();
```
elbalexandre, line 6:
```suggestion
    const result = await runUsersSyncJob();
    return NextResponse.json(result);  
```

apps/notion-api/src/app/api/users/cron/run-sync-user/service.ts
```diff
@@ -0,0 +1,129 @@
...
+}
+
+export const runUsersSyncJob = async () => {
+    try {
+        const job = await prisma.users_Sync_Jobs.findFirst({
+            orderBy: [
+                { priority: 'asc' }, // Order by priority in ascending order
+                { sync_started_at: 'asc' }, // Order by timestamp in ascending order
+            ],
+        });
```
elbalexandre, line 35: This code add condition on `retry_after`. Because it seems like (I am not sure) we can't use postgresql "now()" directly with prisma and `new Date()` isn't safe enough to use, I have to do a raw query here.
```suggestion
        const job: SomeType = await prisma.$queryRaw`SELECT * FROM "users_Sync_Jobs" WHERE "retry_after" < now() ORDER BY "priority" ASC, "sync_started_at" ASC LIMIT 1`;
```

apps/notion-api/src/app/api/authentication/getAccessToken/route.ts
```diff
@@ -0,0 +1,80 @@
...
+    organizationId : string;
+}
+
+type AccessTokenResponse = {
+    workspace_id : string;
+    workspace_name : string; 
+    access_token : string;
+}
+
+export async function POST (req: NextRequest) {
```
elbalexandre, line 19: Could you write a file `service.ts` containing business part (handling notion API call and db transaction) ?
This `route.ts` should only retrieve data from request, call the service, and create a response.

elbalexandre (commented):
Nice improvement from previous PR. The logic and flow are pretty good.
Note that linter has to pass with no error (cf: `pnpm lint`). You might need to install vscode eslint & prettier extension as well.

The services code is also untested, skip that for now, I will ping you when the documentation on testing will be written.

PR #48, opened 2023-12-05
feat(template): enforce inngest usage

elbalexandre:
## Description

- rename `@elba-security/msw-utils` to `@elba-security/test-utils`
- add inngest mock util to `test-utils`
- add `inngest` to required tools in `CONTRIBUTING.md`
- configure `inngest` in **template**
- configure `zod` validation of env variables in **template**
- add an example flow of **authentication** & **users sync** fully tested in **template**
- add `dev:ingest` script in in **template**
- update `database:migrate` script to avoid using npx in **template**

## Additional notes

All the example test are fully functionals.

## Checklist:


template/src/app/api/inngest/route.ts
```diff
@@ -0,0 +1,12 @@
+import { serve } from 'inngest/next';
+import { inngest } from '@/inngest/client';
+import { inngestFunctions } from '@/inngest/functions';
+
+export const runtime = 'edge';
```
proquig, line 5: We might want to mention to delete this line as well if not using edge runtime
Also worth setting the region so our functions are executed close to our database

template/src/app/auth/route.ts
```diff
@@ -0,0 +1,20 @@
+import { RedirectType, redirect } from 'next/navigation';
+import type { NextRequest } from 'next/server';
+import { env } from '@/env';
+import { setupOrganisation } from './service';
+
+/**
+ * This route path can be changed to fit your implementation specificities.
+ */
+export async function GET(request: NextRequest) {
```
proquig, line 9: We should probably add edge runtime config here as well as `export const dynamic = 'force-dynamic';` as we might have cache problem issues in case we use fetch (with the database maybe?)

template/src/env.ts
```diff
@@ -0,0 +1,17 @@
...
+    ELBA_API_BASE_URL: z.string().url(),
+    ELBA_REDIRECT_URL: z.string().url(),
+    ELBA_SOURCE_ID: z.string().uuid(),
+    POSTGRES_URL: z.string().min(1),
+    POSTGRES_HOST: z.string().min(1),
+    POSTGRES_PORT: z.coerce.number().int().positive(),
+    POSTGRES_USER: z.string().min(1),
+    POSTGRES_PASSWORD: z.string().min(1),
+    POSTGRES_DATABASE: z.string().min(1),
+    POSTGRES_PROXY: z.coerce.number().int().positive(),
```
proquig, line 15: I think we should rename this variable `POSTGRES_PROXY_PORT`

template/src/app/auth/service.ts
```diff
@@ -0,0 +1,26 @@
...
+      token,
+    },
+  });
+
+  await inngest.send({
+    name: 'users/sync',
+    data: {
+      isFirstSync: true,
+      organisationId,
+      syncStartedAt: new Date().getTime(),
```
proquig, line 22: We can directly use `Date.now()`

proquig (approved):
Looks good, few minor comments

template/src/app/api/inngest/route.ts
elbalexandre: I will add a comment, regarding the region, it's set in the `vercel.json`.

proquig: The region specified in vercel.json / on the dashboard is the region for lambdas not the edge runtime https://vercel.com/docs/functions/configuring-functions/region#available-configurations

n2ctech: region should be configurable from an env variable as well to be able to deploy both to the US and EU

PR #56, opened 2023-12-14
feat: add webhooks utils and schemas

elbalexandre:
## Description

- add webhook event data in `schemas`
- add `parseWebhookEventData` in `sdk`
- add `validateWebhookRequestSignature` in `sdk` - to be used in a middleware
- update template to includes `ELBA_WEBHOOK_SECRET` env variable
- add example middleware in template using `validateWebhookRequestSignature`

## Additional Notes

N/A

## Checklist:


packages/sdk/src/webhooks/event-data.ts
```diff
@@ -0,0 +1,35 @@
...
+import { ElbaError } from '../error';
+
+const eventDataSchema = {
+  scan_triggered: scanTriggeredWebhookDataSchema,
+  'data_protection.content_requested': dataProtectionContentRequestedDataSchema,
+} as const satisfies Record<string, ZodSchema>;
+
+export type WebhookEvent = keyof typeof eventDataSchema;
+
+export const parseWebhookEventData = <T extends WebhookEvent>(
```
elbalexandre, line 15: I am not sure about this implementation
Pros: 
 - simplify type casting
 - simplify data extraction from search params
 - meaningful errors when testing endpoint locally (if payload is not right)
 
 Type casting implementation:
 pros:
 - No JS
 - straightforward

packages/schemas/src/miscellaneous.ts
```diff
@@ -0,0 +1,5 @@
+import { z } from 'zod';
+
+export const scanTriggeredWebhookDataSchema = z.object({
```
elbalexandre, line 3: There is misunderstand here, data-protection + thrid-party-apps has their own scan-triggered event. There is no general event.

packages/sdk/README.md
```diff
@@ -32,6 +32,51 @@ const elba = new Elba({
...
+Parse webhook event data and validate it against schema.
+
+**JSON payload example:**
+
+```ts
+import { parseWebhookEventData } from 'elba-sdk';
+
+export async function POST(request: NextRequest) {
+  const data: unknown = await request.json();
+  const { organisationId } = parseWebhookEventData('scan_triggered', data);
```
elbalexandre, line 66:
```suggestion
  const { organisationId } = parseWebhookEventData('data_protection.scan_triggered', data);
```

packages/sdk/src/webhooks/request-signature.ts
```diff
@@ -23,13 +70,13 @@
  * console.log("Is the signature valid?, isValid ? "Yes" : "No");
  * ```
  */
-export async function isValidWebhookElbaSignature(
+async function validateSignature(
   secret: string,
-  payload: Record<string, unknown>,
-  elbaSignatureFromHeader: string
+  payload: Record<string, unknown> | undefined,
```
proquig, line 75: The more I see this, the more I'm convinced that it's not the right way to handle it. First, what if somehow elba sends prettified json, then the signature wouldn't match, same if somehow the order of the generated json string is not the same between elba and the integration.
For me we should just sign the raw text and verify it matches the elba signature. One benefit is that we're not stuck only supporting signing json.
This is what is happening for slack for example, as the signature also contains the timestamp of the request to avoid replay attacks

elbalexandre: We are generating signature from a record as well, so it make sense to check it against a record. It should always works.

We can use also use `request.text()` to verify it. It could be easier to check search params for the GET request as well.

elbalexandre: @raulelba If it's ok for you, I am going to change both the signature (in elba) and verification here.

proquig: The same has been done on elba side

elbalexandre: > for the request will always be a JSON object

Not for retrieving the data protection object content, it will be the search params if we keep this API as it is.

proquig (approved):

packages/sdk/src/webhooks/request-signature.ts
proquig: On the DP side the fetch data protection object content webhook was not not signed by elba and was just a GET with query params. The PR to update this is [here](https://github.com/elba-security/elba/pull/3159)

packages/sdk/src/webhooks/request-signature.ts
```diff
@@ -1,3 +1,44 @@
...
+
+  if (!isSignatureValid) {
+    throw new ElbaError('Could not validate elba signature from webhook request', {
+      request,
+    });
+  }
+};
+
+const getRequestPayload = async (request: Request) => {
+  if (['POST', 'PATCH', 'PUT'].includes(request.method)) {
```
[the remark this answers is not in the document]
elbalexandre: Only `POST` `PATCH` & `PUT` can have a body. My guess is if we need to had some other method we will have no update to do here. But your comment seems legit, doing only POST for webhook seems legit and also futur proof. I am going to remove that.

PR #71, opened 2024-01-08
chore: add utils package

proquig:
## Description

This adds a utils package that contains a collection of utils. For now it only contains crypto utils.

## Checklist:

### DIFF
--- a/packages/utils/src/crypto.ts
+++ b/packages/utils/src/crypto.ts
@@ -1,4 +1,4 @@
-const arrayBufferFromHexString = (hexString: string) => {
+export const arrayBufferFromHexString = (hexString: string) => {
   const matches = hexString.match(/[0-9a-f]{2}/gi);
   if (!matches || hexString.length !== matches.length * 2) {
     throw new Error('Invalid hex string');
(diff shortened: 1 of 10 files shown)


packages/utils/src/crypto.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  const bytes = Uint8Array.from(matches.map((hex) => parseInt(hex, 16)));
+  return bytes.buffer;
+};
+
+const hexStringFromArrayBuffer = (buffer: ArrayBuffer) => {
+  return [...new Uint8Array(buffer)].map((byte) => byte.toString(16).padStart(2, '0')).join('');
+};
+
+export const encryptAES256GCM = async ({ data, keyHex }: { data: string; keyHex: string }) => {
```
elbalexandre, line 15: minor, but IMO `encryptAES256GCM(data: string, key: string)` is in this case a better signature for callee and declaration.

Do we really need to precise `keyHex` ? looks odd as `userIdUuid`.

elbalexandre (commented):
Can we add test with vitest edge-runtime env ?
We might want to rename `encryptAES256GCM` & `decryptAES256GCM`. IMO it's a bit too barbaric. Perhaps `encryptText` & `decryptText` could be fine.

elbalexandre (approved):

PR #72, opened 2024-01-08
[Draft] Teams Integeration

naveed-niazi:
## Description

## Checklist:


apps/teams/src/app/install/route.ts
```diff
@@ -0,0 +1,13 @@
...
+import { redirect } from 'next/navigation';
+import { type NextRequest } from 'next/server';
+import { env } from '@/env';
+
+export const dynamic = 'force-dynamic';
+
+export function GET(request: NextRequest) {
+  cookies().set('organisation_id', env.ELBA_ORGANIZATION_ID);
+
+  let redirectUrl = `${env.AZURE_AUTH_URL}?response_type=code&scope=openid+profile+email+offline_access&client_id=${env.AZURE_AD_CLIENT_ID}&&redirect_uri=${env.AZURE_AUTH_REDIRECT_URL}`;
```
elbalexandre, line 11: There is a typo here `&&redirect_uri`.

We might prefer to use the `URL` object and append search params, it's safer.

Additionally, we prefer that we use `let` keyword only when the variable changes `eslint[prefer-const](https://eslint.org/docs/latest/rules/prefer-const)`

apps/teams/src/app/auth/route.ts
```diff
@@ -0,0 +1,30 @@
+import { RedirectType, redirect } from 'next/navigation';
```
elbalexandre, line 1: If you are not going to use this router (as `setup`  exists) you might prefer to delete it right away.

apps/teams/src/app/setup/route.ts
```diff
@@ -0,0 +1,51 @@
+import { RedirectType, redirect } from 'next/navigation';
+import { isRedirectError } from 'next/dist/client/components/redirect';
+import type { NextRequest } from 'next/server';
+import { env } from '@/env';
+import { setupAdmin, setupOrganisation } from './service';
+import axios from 'axios';
```
elbalexandre, line 6: We should `fetch` instead, `axios` is not compatible with edge-runtime.

apps/teams/src/app/setup/route.ts
```diff
@@ -0,0 +1,51 @@
...
+export const dynamic = 'force-dynamic';
+
```
elbalexandre, line 8: the environment should be edge-runtime

apps/teams/src/app/setup/route.ts
```diff
@@ -0,0 +1,51 @@
...
+    const result = await axios.post(URL, requestBody, {
+      headers: { 'content-type': 'application/x-www-form-urlencoded' },
+    });
+
+    const { access_token, refresh_token } = result.data;
+    if (!access_token || !refresh_token) {
+      console.log('Error: Either access_token or refresh_token is missing');
+      redirect(`${env.ELBA_REDIRECT_URL}?error=true`, RedirectType.replace);
+    }
```
elbalexandre, line 32: All this code can be moved in `/connectors/auth` in a dedicated function.
The URL of the API should be hardcoded (it should in an env variable).

apps/teams/src/app/setup/route.ts
```diff
@@ -0,0 +1,51 @@
+import { RedirectType, redirect } from 'next/navigation';
+import { isRedirectError } from 'next/dist/client/components/redirect';
+import type { NextRequest } from 'next/server';
+import { env } from '@/env';
+import { setupAdmin, setupOrganisation } from './service';
+import axios from 'axios';
+export const dynamic = 'force-dynamic';
+
+export async function GET(request: NextRequest) {
+  try {
```
elbalexandre, line 10: this function should only handle `request` data extraction + `response` creation. Everything else should be a service.

apps/teams/src/app/setup/service.ts
```diff
@@ -0,0 +1,67 @@
...
+      console.log('Error: ', err);
+      redirect(
+        `${env.ELBA_REDIRECT_URL}?source=${env.ELBA_SOURCE_ID}&error=${
+          err?.response?.status ?? 500
+        }`,
+        RedirectType.replace
+      );
+    }
+  }
+};
```
elbalexandre, line 33: What is the purpose of storing this admin ?

apps/teams/src/app/setup/service.ts
```diff
@@ -0,0 +1,67 @@
...
+  access_token: string,
+  refresh_token: string
+) => {
+  const [organisation] = await db
+    .insert(Organisation)
+    .values({
+      id: organisationId,
+      access_token: access_token,
+      refresh_token: refresh_token,
+      createdAt: new Date(),
```
elbalexandre, line 46: the `createdAt` date is automatically filled by postgres, we don't need to precise it.

apps/teams/src/app/setup/service.ts
```diff
@@ -0,0 +1,67 @@
+import { env } from '@/env';
```
elbalexandre, line 1: A test file is missing for this service.

elbalexandre (commented):
The flow is correct, but the code does not respect separation of concerns.

You might want to install eslint + prettier extension as this PR do not pass lint stage. Note that if you rebase your branch on staging this stage will be done while commiting.

apps/teams/.eslintrc.js
```diff
@@ -0,0 +1,17 @@
+const { resolve } = require('node:path');
+
+const project = resolve(__dirname, './tsconfig.json');
+
+module.exports = {
+  extends: ['@elba-security/eslint-config-custom/next', 'plugin:@typescript-eslint/recommended'],
```
elbalexandre, line 6:
```suggestion
  extends: ['@elba-security/eslint-config-custom/next'],
```
It's better to stick to the configuration we are providing for now. If in your opinion it's lake of rules or some might be changed, feels free to open an issue or discute it with me in pm.

apps/teams/src/connectors/auth.ts
```diff
@@ -0,0 +1,24 @@
...
+  });
+
+  const response = await fetch(env.AZURE_TOKENS_URL, {
+    method: 'POST',
+    headers: { 'content-type': 'application/x-www-form-urlencoded' },
+    body: requestBody.toString(),
+  });
+
+  const result = (await response.json()) as TokenResponseData;
+
```
elbalexandre, line 22: Could you throw a `TeamError` (you can rename `MySaasError`) containing the response when `response.ok` is falsy ?

apps/teams/src/connectors/auth.test.ts
```diff
@@ -0,0 +1,44 @@
+/* eslint-disable @typescript-eslint/no-unsafe-call -- test conveniency */
```
elbalexandre, line 1: Could you implement the tests here ?

apps/teams/src/app/setup/route.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (!code || !organisationId) {
+    throw new Error('Code or organisationId is missing.');
+  }
+
+  return { code, organisationId };
+}
+
+function redirectTo(url: string): void {
+  redirect(url, RedirectType.replace);
+}
```
elbalexandre, line 34: IMO, we can remove this implementation:
- it's used once
- it's overwriting nextjs API which should be known by others devs

apps/teams/src/app/setup/route.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export async function GET(request: NextRequest): Promise<void> {
+  let redirectUrl = '';
+  try {
+    const { code, organisationId } = validateRequest(request);
+    const tokens = await getTokens(code);
+    await setupOrganisation(organisationId, tokens.accessToken, tokens.refreshToken);
+    redirectUrl = `${env.ELBA_REDIRECT_URL}?source=${env.ELBA_SOURCE_ID}&success=true`;
+  } catch (error: unknown) {
+    redirectUrl = `${env.ELBA_REDIRECT_URL}?source=${env.ELBA_SOURCE_ID}&error=true`;
```
elbalexandre, line 15: Sorry, we did not document this and the template might be out of date, but we are now supporting error code on elba.

```suggestion
   if (error instanceof TeamsError && error.response.status === 401) {
     redirectUrl = `${env.ELBA_REDIRECT_URL}?source=${env.ELBA_SOURCE_ID}&error=unauthorized`;
   } else {
     redirectUrl = `${env.ELBA_REDIRECT_URL}?source=${env.ELBA_SOURCE_ID}&error=internal_error`;
   }
```

elbalexandre (commented):
Great work, almost there.
Please, for next PRs, could you add a description. I am assuming this PR cover only app installation feature.

apps/teams/.env.local.example
```diff
@@ -0,0 +1,12 @@
+ELBA_SOURCE_ID="0442f541-45d2-487a-9e4b-de03ce4c559e"
```
elbalexandre, line 1: This should be renamed as `.env.local`

apps/teams/src/app/setup/route.ts
naveed-niazi: Sure thing, I will just directly redirect

apps/teams/.env.local.example
naveed-niazi: Updated it to .env.local

elbalexandre: It still there, you can remove this one.

naveed-niazi: I haven't pushed all the code I will put it for review to you when all of it is complete. Thanks for the detailed review

PR #83, opened 2024-01-16
feat(dbx): data protection

athielba:
## Description

Complete data protection integration added in this PR

## Checklist:


apps/dropbox/src/connectors/dropbox/dbx-files.ts
```diff
@@ -0,0 +1,365 @@
...
+        return acc;
+      },
+      {
+        folders: [],
+        files: [],
+      }
+    );
+
+    const [foldersPermissions, foldersMetadata, filesPermissions, filesMetadata] =
+      await Promise.all([
```
elbalexandre, line 321: I hope it will not timeout or rate limit, that's a bunch of request to do.

apps/dropbox/src/connectors/utils/format-shared-links.ts
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    const linkObject = {
+      id,
+      url,
+      linkAccessLevel: link.link_permissions.link_access_level?.['.tag'] ?? 'viewer',
+      pathLower: path_lower!,
+    };
+
+    return [...acc, linkObject];
```
elbalexandre, line 28:
```suggestion
    acc.push(linkObject)
    return acc;
```
https://www.measurethat.net/Benchmarks/Show/29740/0/reduce-array-destructuring-vs-push

apps/dropbox/src/database/schema.ts
```diff
@@ -10,3 +10,20 @@ export const organisations = pgTable('organisations', {
...
+    id: text('id').notNull(),
+    url: text('url').notNull(),
+    organisationId: uuid('organisation_id').notNull(),
+    teamMemberId: text('team_member_id').notNull(),
+    linkAccessLevel: text('link_access_level').notNull(),
+    pathLower: text('path_lower').notNull(),
+  },
+  (table) => {
+    return {
+      pk: primaryKey({ columns: [table.url, table.pathLower] }),
```
elbalexandre, line 26: Why not use id as pk ? it's not unique across all dbx organisations ?

apps/dropbox/src/inngest/functions/data-protection/create-sync-folders-and-files.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  await step.run('delete-objects', async () => {
+    await elba.dataProtection.deleteObjects({
+      syncedBefore: new Date(syncStartedAt).toISOString(),
+    });
+  });
+};
+
+export const createPathSyncJobs = inngest.createFunction(
```
elbalexandre, line 77:
```suggestion
export const createPathSyncJobs = inngest.createFunction(
```
The syncjobs part could be confusing and do not really match the event triggering the function. Do you have any other name in your naming hat ?

apps/dropbox/src/inngest/functions/data-protection/create-sync-shared-links.ts
```diff
@@ -0,0 +1,115 @@
...
+    name: 'dropbox/data_protection.folder_and_files.create.sync_page.requested',
+    data: {
+      organisationId,
+      syncStartedAt,
+      isFirstSync,
+    },
+  });
+};
+
+export const createSharedLinkSyncJobs = inngest.createFunction(
```
elbalexandre, line 101:
```suggestion
export const createSharedLinkSyncJobs = inngest.createFunction(
```
Same here

apps/dropbox/src/inngest/functions/data-protection/create-sync-shared-links.ts
```diff
@@ -0,0 +1,115 @@
...
+        teamMemberId: profile.team_member_id,
+        isPersonal: true,
+      },
+    ];
+  });
+
+  if (team.members.length > 0) {
+    const eventsToWait = sharedLinkJobs.map(
+      async (sharedLinkJob) =>
+        await step.waitForEvent(`dropbox-wait-sync-shared-links`, {
```
elbalexandre, line 62:
```suggestion
        await step.waitForEvent(`wait-sync-shared-links`, {
```

I don't think we need to suffix function step as on inngest we can only see them in a function run.

elbalexandre (commented):
Let's be cautious with"job" while naming tests and inngest functions as it's not in the vocabulary of inngest. (function run could replace it in the test)

apps/dropbox/src/connectors/dropbox/dbx-files.ts
athielba: I hope we won't need to do more right now. Let's see how things go first. If it doesn't work out, I'll set up some events to help us out. 🤞🏼

apps/dropbox/src/database/schema.ts
athielba: No, a file can have multiple shared links, which has the same file or folder id

apps/dropbox/src/inngest/functions/data-protection/create-sync-shared-links.ts
athielba: I would like to keep it since I used this pattern in the all Dbx integration

apps/dropbox/src/database/schema.ts
elbalexandre: The id of a shared link is not unique then ?

apps/dropbox/src/inngest/functions/data-protection/create-sync-shared-links.ts
elbalexandre: A search and replace can changes this for every occurence in a minute.
I think keeping integration consistent is a must have.

apps/dropbox/src/inngest/functions/data-protection/create-sync-folders-and-files.ts
```diff
@@ -0,0 +1,91 @@
...
+    accessToken: token,
+  });
+
+  const elba = getElba({
+    organisationId,
+    region,
+  });
+
+  const team = await step.run('fetch-users', async () => {
+    return dbxUsers.fetchUsers(cursor);
```
elbalexandre, line 36: IMO if we need the users for DP we should store them in the db and update them at each users sync

apps/dropbox/src/inngest/functions/data-protection/create-sync-folders-and-files.ts
elbalexandre: This naming is still an issue. What does the function is unclear ? After looking at the code it's seems to start the folders & files sync, so why not using `start-folders-and-files-sync` ?

elbalexandre (commented):
IMO, the naming of the events is messy:
- what is the entrypoint for entity sync ?
- creating a sync (sync is a short term of synchronise) is a strange term
- no differentiator between page sync & sync start

I suggest to use the prefix `start`, like `start-sync-folders-and-files` or `start-folders-and-files-sync`.

We should be really sure of the event naming as it will be really hard to refactor afterward + handling migration.

In addition, doing a users sync inside the data-protection sync is IMO not a good idea:
cons:
- if their is some new/deleted users it will not create issues for them (so syncing their files is useless)
- it's taking rate-limit bandwidth for data we have already retrieved previously
pros:
- less code

I suggest storing the users in a table (during users sync) and retrieve them when needed.

athielba:
@elbalexandre ,
1.  I agree withe file naming convention, since  DBX need to create shared sync & folder file sync events, the files and the functions have been updates
2. I don't understand what do you mean by user sync inside `data-protection`,  the shared-link sync and file-folder sync event will be triggered based on the current user details fetched from Dbx since we need the lates root id of the user, also we ignore the use if they delete or invited, we only  fetch the user who have the state of [`'active', 'suspended']`
 if your concern it is not relate to the the above please reference the code below

elbalexandre:
> @elbalexandre ,
> 
> 1. I agree withe file naming convention, since  DBX need to create shared sync & folder file sync events, the files and the functions have been updates
> 2. I don't understand what do you mean by user sync inside `data-protection`,  the shared-link sync and file-folder sync event will be triggered based on the current user details fetched from Dbx since we need the lates root id of the user, also we ignore the use if they delete or invited, we only  fetch the user who have the state of [`'active', 'suspended']`
>    if your concern it is not relate to the the above please reference the code below

Why do we need to fetch the users from dropbox when we can store them at the users syncs ?
Why do we paginate users sync but not the step retrieving the users in data-protection ?

athielba:
@elbalexandre  
I believe you missed the  code where the pagination is  created
1.  [Start shared link](https://github.com/elba-security/elba-security/blob/dfe30bd667f393ba1c4418de33664f5bcb173b2d/apps/dropbox/src/inngest/functions/data-protection/start-shared-links-sync.ts#L80-L81)
2. [Start folde and files](https://github.com/elba-security/elba-security/blob/dfe30bd667f393ba1c4418de33664f5bcb173b2d/apps/dropbox/src/inngest/functions/data-protection/start-folders-and-files-sync.ts#L60-L61)

elbalexandre (approved):

PR #88, opened 2024-01-18
fix: elba signature verification

proquig:
## Description

This fixes a bug where validating elba signature was not working due to the fact it was trying to JSON stringify a string, adding unecessary quotes wrapping the string and failing to verify correctly that the signatures were matching.

## Checklist:

### DIFF
--- a/packages/sdk/src/webhooks/request-signature.test.ts
+++ b/packages/sdk/src/webhooks/request-signature.test.ts
@@ -5,13 +5,13 @@ import { validateWebhookRequestSignature } from './request-signature';
 describe('validateWebhookRequestSignature', () => {
   it('should succeed when the request signature is valid', async () => {
     const secret = 'test-secret';
-    const payload = '{ "data": "example" }';
+    const payload = '{"data":"example"}';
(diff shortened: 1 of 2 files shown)


packages/sdk/src/webhooks/request-signature.test.ts
```diff
@@ -5,27 +5,27 @@ import { validateWebhookRequestSignature } from './request-signature';
...
-    const payload = '{ "data": "example" }';
+    const payload = '{"data":"example"}';
 
     const signature = 'invalid-signature';
     const request = new Request(new URL('http://foo.bar'), {
       method: 'post',
-      body: JSON.stringify(payload),
-      headers: { 'X-Elba-Signature': signature },
+      body: payload,
+      headers: { 'X-elba-Signature': signature },
```
athielba, line 28: can we use uppercase or lowercase for all `x-elba-signature` ?

proquig: It's case insensitive when we retrieve the header, I just updated it to match what we had on elba side

PR #97, opened 2024-01-25
feat(tpa-webhook-validation): add tpa refresh & delete object request webhooks validation

athielba:
## Description

Include webhook validation for 
1. third party app refresh object
2. third party app delete object

## Checklist:

### DIFF
--- a/packages/schemas/src/third-party-apps.ts
+++ b/packages/schemas/src/third-party-apps.ts
@@ -41,3 +41,18 @@ export type DeleteThirdPartyApps = zInfer<typeof deleteThirdPartyAppsSchema>;
 export const thirdPartyAppsScanTriggeredWebhookDataSchema = z.object({
   organisationId: z.string().uuid(),
 });
+
+export const thirdPartyAppsRefreshObjectRequestedWebhookDataSchema = z.object({
+  organisationId: z.string().uuid(),
+  userId: z.string(),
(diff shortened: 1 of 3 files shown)


packages/schemas/src/third-party-apps.ts
```diff
@@ -41,3 +41,18 @@ export type DeleteThirdPartyApps = zInfer<typeof deleteThirdPartyAppsSchema>;
...
+  userId: z.string(),
+  appId: z.string(),
+  metadata: z.any()
+});
+
+export const thirdPartyAppsDeleteObjectRequestedWebhookDataSchema = z.object({
+  organisationId: z.string().uuid(),
+  userId: z.string(),
+  appId: z.string(),
+  metadata: z.any()
```
elbalexandre, line 56: There is `jsonSchema` in `common.ts` that is more safe than any.

packages/sdk/src/webhooks/event-data.test.ts
```diff
@@ -3,28 +3,64 @@ import { ElbaError } from '../error';
 import { parseWebhookEventData } from './event-data';
```
elbalexandre, line 4: no need to changes that, we are not going to test each webhook event here, we can trust zod for that. The two existing test was for covering general behavior.

elbalexandre (approved):

PR #98, opened 2024-01-25
docs(template): update inngest event names

elbalexandre:
## Description

- update template inngest event name

## Additional Notes

n/a

## Checklist:

### DIFF
--- a/template/src/inngest/functions/users/sync-users-page.test.ts
+++ b/template/src/inngest/functions/users/sync-users-page.test.ts
@@ -28,7 +28,7 @@ const users: usersConnector.MySaasUser[] = Array.from({ length: 5 }, (_, i) => (
   email: `username-${i}@foo.bar`,
 }));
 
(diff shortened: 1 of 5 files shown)


template/src/inngest/functions/users/sync-users-page.test.ts
```diff
@@ -28,7 +28,7 @@ const users: usersConnector.MySaasUser[] = Array.from({ length: 5 }, (_, i) => (
   email: `username-${i}@foo.bar`,
 }));
 
-const setup = createInngestFunctionMock(syncUsersPage, 'users/sync_page.triggered');
+const setup = createInngestFunctionMock(syncUsersPage, '{SaaS}/users.page_sync.requested');
```
athielba, line 31: IMO, since it is not doc we can directly use a source name(`github`) for instance , and leave a  simple comment that  to say source will be differ be based on the integration

PR #106, opened 2024-02-01
chore: enable turbo cache on github workflows

elbalexandre:
## Description

- enable turbo cache on github workflows

## Additional Notes

N/a

## Checklist:

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -30,6 +30,15 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Cache TurboRepo
+        uses: actions/cache@v3
+        id: node-modules-cache
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-quality_node_modules-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-quality_node_modules-
+
       - name: Lint apps and packages
         run: pnpm lint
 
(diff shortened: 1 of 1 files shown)


.github/workflows/quality.yml
```diff
@@ -30,6 +30,15 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Cache TurboRepo
```
ValentinH, line 33: What's the goal for doing this?
It doesn't seem to be Turborepo specific but instead just caching node modules. In that case, this should be done before the "Instal deps" step to really be useful.

If you want to use Turbo remote cache, we should instead setup the right env variables to use our Vercel remote cache plan. See https://turbo.build/repo/docs/ci/github-actions

.github/workflows/quality.yml
```diff
@@ -14,6 +14,9 @@ jobs:
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    env:
+      TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
+      TURBO_TEAM: ${{ vars.TURBO_TEAM }}
```
ValentinH, line 19: where is `vars.TURBO_TEAM` defined?

elbalexandre: In Github settings

ValentinH (approved):

.github/workflows/quality.yml
ValentinH: Nice didn't know that. better than using a secret indeed!

PR #117, opened 2024-02-08
feat(api-client): add verbose mode

elbalexandre:
## Description

- add `-v` argument to `start` script that enable verbose mode
- when verbose mode is on every request bodies are logged

## Additional Notes

N/A

## Checklist:

- [X] I have commented my code, particularly in hard-to-understand areas.

### DIFF
--- a/packages/api-client/package.json
+++ b/packages/api-client/package.json
@@ -6,7 +6,8 @@
   "scripts": {
     "lint": "eslint server.ts src",
     "lint:fix": "pnpm lint --fix",
-    "start": "tsx watch server.ts -v"
+    "start": "tsx watch server.ts -v",
+    "start:quiet": "tsx watch server.ts"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


athielba (approved):

elbalexandre:
> All looks good to me, but why would we add the `-v` argument to the script instead of always logging the logs? Or setting it in a constant?

My guess it's can be handy to start it silently when it's bloating the log. I could add a `start:quiet` in the script.

elbalexandre:
> > > All looks good to me, but why would we add the `-v` argument to the script instead of always logging the logs? Or setting it in a constant?
> > 
> > 
> > My guess it's can be handy to start it silently when it's bloating the log. I could add a `start:quiet` in the script.
> 
> Fair enough. I figured as much, but IMO if that's necessary it always can be done later. Not a bad thing that you did it this way anyways.

I have added it.

PR #129, opened 2024-02-13
feat(segment): add install & users sync

alifarooq96:
## Description

This PR includes integration for Segment

## Additional Notes

Here's an example URL after running the project which would take you to the UI for Segment: http://localhost:4000/install?organisation_id=0442f541-45d2-487a-9e4b-de03ce4c559e&region=pakistan

## Checklist:


apps/segment/src/connectors/users.test.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const users: SegmentUser[] = [
+  {
+    id: `d2oLAmTuScRdhhAnrJJwN3`,
+    name: `Ali Farooq`,
+    email: `ali@alfabolt.com`,
+  },
+];
+
+describe('getUsers', () => {
```
elbalexandre, line 27: We also need to check the nextPage property returned here.
I suggest to add a test that check nextPage.next is null when the end of the list is reached. This implies to extends the msw handler to handle this case.

apps/segment/src/app/install/action.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  try {
+    // handling await
+    await registerOrganisation({ organisationId, token, region });
+    return {
+      redirectUrl: `${env.ELBA_REDIRECT_URL}?source_id=${env.ELBA_SOURCE_ID}&success=true`,
+    };
+  } catch (error) {
+    logger.warn('Could not register organisation', { error });
+    if (error instanceof SegmentError && error.response && error.response.status === 401) {
```
elbalexandre, line 54:
```suggestion
    if (error instanceof SegmentError && error.response?.status === 401) {
```

apps/segment/src/connectors/commons/error.ts
```diff
@@ -0,0 +1,18 @@
+type SegmentErrorOptions = { response?: Response; request?: Request };
+
+export class SegmentError extends Error {
+  statusCode?: number;
```
elbalexandre, line 4:
```suggestion
```
It's already in the response, and I think it's not used anyway.

apps/segment/src/app/install/service.ts
```diff
@@ -0,0 +1,43 @@
...
+import { Organisation } from '@/database/schema';
+import { inngest } from '@/inngest/client';
+
+type SetupOrganisationParams = {
+  organisationId: string;
+  token: string;
+  region: string;
+};
+
+export const registerOrganisation = async ({
```
elbalexandre, line 11: In this function we also need to make sure the data provided in the form by the user is correct.

We could do a call to this api (echo) https://docs.segmentapis.com/tag/Testing to make sure the token is right.
If the response is 401 the catch block in the action will redirect with unauthorized, which is fine.

apps/segment/src/connectors/users.test.ts
alifarooq96: Done

apps/segment/src/app/install/action.tsx
alifarooq96: Updated

apps/segment/src/connectors/commons/error.ts
alifarooq96: Updated

apps/segment/src/connectors/users.ts
```diff
@@ -0,0 +1,32 @@
...
+};
+
+type GetUsersResponseData = { users: SegmentUser[]; nextPage: Pagination };
+
+export const getUsers = async (token: string, page: number | null) => {
+  // Setting default value for pagination.count
+  const count = 200;
+
+  const response = await fetch(
+    `https://api.segmentapis.com/users?pagination.count=${count}&pagination.cursor=${page}`,
```
elbalexandre, line 23:
```suggestion
    `https://api.segmentapis.com/users?pagination.count=${env.USERS_SYNC_BATCH_SIZE}&pagination.cursor=${page}`,
```

and in `env.ts` -> `USERS_SYNC_BATCH_SIZE: z.coerce.number().int().positive().default(200)`

apps/segment/src/connectors/users.test.ts
```diff
@@ -0,0 +1,55 @@
...
+const validToken = env.SEGMENT_API_TOKEN;
+
+describe('getUsers', () => {
+  beforeEach(() => {
+    server.use(
+      http.get('https://api.segmentapis.com/users', ({ request }) => {
+        if (request.headers.get('Authorization') !== `Bearer ${validToken}`) {
+          return new Response(undefined, { status: 401 });
+        }
+        return new Response(JSON.stringify({ users, nextPage: pagination }), { status: 200 });
```
elbalexandre, line 32:
```suggestion
const lastCursor = 'last-cursor'
const nextCursor = 'next-cursor'
const previousCursor = 'previous-cursor'
// ...
        const url = new URL(request.url);
        const pagination = url.searchParams.get('pagination');
        const cursor = pagination.cursor
        return new Response(JSON.stringify({
          users, nextPage: { ...pagination, next: cursor === lastCursor ? null : nextCursor, previous: previousCursor }
        }), { status: 200 });
```

Could you use this for the mock, and test that the a page that don't use last cursor get a next cursor + page that use last cursor get a null next cursor.

apps/segment/src/connectors/auth.ts
```diff
@@ -0,0 +1,7 @@
+export const validateToken = async (token: string) => {
+  const response = await fetch(`https://api.segmentapis.com/echo?message=valid`, {
+    headers: { Authorization: `Bearer ${token}` },
+  });
+
+  return response.status;
```
elbalexandre, line 6:
```suggestion
if (!response.ok) {
  throw new SegmentError('Could not validate token', { response })
}
```

No need to returns true, if the function don't throw, the token is valid.

apps/segment/src/app/install/action.tsx
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  const organisationId = result.data.organisationId;
+  const region = result.data.region;
+  const token = result.data.token;
+
+  try {
+    // handling await
+    const statusCode = await validateToken(token);
+    if (statusCode === 200) {
```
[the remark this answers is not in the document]
elbalexandre: No need to check for the response status. Just call validateToken in the service+
update the test and add a test case: `should throw when the token is invalid`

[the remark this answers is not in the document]
awan7676: Tests related to validateToken are written in a separate file auth.test.ts. 
In Service.test we are just creating a mock for validateToken to bypass the request as it's already being tested and focus on registerOrganisation functionality itself.
This will be visible in the next commit.

apps/segment/src/connectors/users.ts
awan7676: Done

apps/segment/src/connectors/users.test.ts
awan7676: Done

apps/segment/src/connectors/auth.ts
awan7676: Done

apps/segment/src/app/install/action.tsx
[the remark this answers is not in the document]
awan7676: Tests related to validateToken are written in a separate file named auth.test.ts.
We are creating a mock for validateToken in service.test.ts and focusing on testing the behavior of registerOrganisation function.

apps/segment/src/connectors/auth.ts
elbalexandre: I still seeing the response.status

apps/segment/src/app/install/action.tsx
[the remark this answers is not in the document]
elbalexandre: I am still seeing the validateToken call here instead being in the service.ts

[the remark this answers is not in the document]
awan7676: Just made a commit. You can check it now.

elbalexandre (approved):

PR #135, opened 2024-02-14
feat(dbx): include uninstall organisation event

athielba:
## Description

Include uninstall organisation event 

## Checklist:


apps/dropbox/src/inngest/functions/organisations/remove-organisation.ts
```diff
@@ -0,0 +1,50 @@
...
+import { env } from '@/env';
+import { inngest } from '../../client';
+import { organisations } from '@/database';
+
+export const removeOrganisation = inngest.createFunction(
+  {
+    id: 'dropbox-remove-organisation',
+    priority: {
+      run: '600',
+    },
```
elbalexandre, line 14:
```suggestion
```
I don't know why I put this here: priority is applied only to similar function, perhaps I though it was global to the app.

This event is going to be the same for most (or every) integration, perhaps we could factorize it.

elbalexandre (approved):

PR #137, opened 2024-02-15
fix(dbx-token-refresh): fix cancel refresh token on multiple app install request

athielba:
## Description

 fix cancel refresh token on multiple app install request

## Checklist:


elbalexandre (approved):
Are the TPA sync canceled by the uninstallation event ?

I think on elba we should re-trigger a scan when the source is connected and first_scan_status is started, otherwise if the user remove the app while the first scan it will not be retriggered.

athielba:
> Are the TPA sync canceled by the uninstallation event ?
> 
> I think on elba we should re-trigger a scan when the source is connected and first_scan_status is started, otherwise if the user remove the app while the first scan it will not be retriggered.

Tnx for point me, I will update it

PR #165, opened 2024-03-09
Calendly Integration

awan7676:


apps/calendly/src/app/api/webhooks/elba/users/delete/route.ts
```diff
@@ -0,0 +1,18 @@
+import { NextResponse } from 'next/server';
+import { parseWebhookEventData } from '@elba-security/sdk';
+import { deleteUserRequest } from './service';
+
+export async function DELETE(request: Request) {
+  const data: unknown = await request.json();
+
+  const { id, organisationId } = parseWebhookEventData('users.delete_user_requested', data);
+
+  const userId: string = id as string; // Explicitly type id as string
```
elbalexandre, line 10: What was the type of `id` ? The `parseWebhookEventData` should have returned the type `string` for id (it could be bug).

awan7676: It was showing some kind of warning, so i explicitly declared it as a string.

apps/calendly/src/connectors/auth.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  if (!response.ok) {
+    throw new CalendlyError('Failed to fetch', { response });
+  }
+  const data = (await response.json()) as GetTokenResponseData;
+  const tokenResponse = {
+    accessToken: data.access_token,
+    refreshToken: data.refresh_token,
+    expiresIn: data.expires_in,
+    organization: data.organization,
```
elbalexandre, line 32:
```suggestion
    organizationId: data.organization,
```
Let's call it id if it's an id.

apps/calendly/src/inngest/functions/users/delete-user.ts
```diff
@@ -0,0 +1,47 @@
...
+        })
+        .from(Organisation)
+        .where(eq(Organisation.id, organisationId));
+      if (!organisation) {
+        throw new NonRetriableError(`Could not retrieve organisation with id=${organisationId}`);
+      }
+      return organisation.accessToken;
+    });
+
+    if (accessToken) {
```
elbalexandre, line 37:
```suggestion
```
`accessToken` should always be a string as in the db schema it's `.notNull()`

apps/calendly/src/inngest/functions/users/delete-user.ts
```diff
@@ -0,0 +1,47 @@
...
+      await step.run('delete-user', async () => {
+        await deleteUser(accessToken, id);
```
elbalexandre, line 39: The `id` is used here instead of the endpoint uri.
IMO in the in `deleteUser` function we should construct the delete endpoint uri from organizationUri.

apps/calendly/src/database/schema.ts
```diff
@@ -0,0 +1,11 @@
+import { uuid, text, timestamp, pgTable } from 'drizzle-orm/pg-core';
+
+export const Organisation = pgTable('organisation', {
+  id: uuid('id').notNull().primaryKey(),
+  accessToken: text('access_token').notNull(),
+  refreshToken: text('refresh_token').notNull(),
+  organizationUri: text('organization_uri').notNull(),
```
elbalexandre, line 7:
```suggestion
  organizationId: text('organization_id').notNull(),
```

Is it an id or a uri ?

PR #171, opened 2024-03-13
feat(webhooks): initial MS teams webhook

yuriiyf:
## Description

Initial webhook logic implementation for MS teams to keep up in sync messages and replies. There is still a lot of work, so don’t waste too much time reviewing it.

1. I am still testing all scenarios, so will update code after that, eventually
2. Implement proper channel deletion, message and reply updated event handlers
3. Write a tests for all functions

## Checklist:


apps/teams/src/inngest/functions/subscribtions/subscription-to-channels.ts
```diff
@@ -0,0 +1,60 @@
+import { addDays } from 'date-fns';
```
elbalexandre, line 1: The folder is named `subcribtions` instead of `subscriptions`

apps/teams/src/inngest/functions/subscribtions/subscription-to-channels.ts
```diff
@@ -0,0 +1,60 @@
...
+      .from(organisationsTable)
+      .where(eq(organisationsTable.id, organisationId));
+
+    if (!organisation) {
+      throw new NonRetriableError(`Could not retrieve organisation with id=${organisationId}`);
+    }
+
+    const token = await decrypt(organisation.token);
+
+    await fetch(`${env.MICROSOFT_API_URL}/subscriptions`, {
```
elbalexandre, line 45: If this could be in a connector: throwing TeamsError as well and be individually tested it could be nice

apps/teams/src/app/api/lifecycleNotifications/service.ts
```diff
@@ -0,0 +1,43 @@
+import { z } from 'zod';
```
elbalexandre, line 1: folder name should be in `kebab-case` if microsoft allows it.

elbalexandre (commented):
The implementation looks solid, but it preferable to test everything (connector + service + inngest function) before merging.
Anyway, still a great job !

PR #178, opened 2024-03-18
complete aircall

mbliakharskyi:
## Description

## Checklist:


apps/aircall/src/connectors/users.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  const response = await fetch(url.toString(), {
+    method: 'DELETE',
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: `Bearer ${token}`,
+    }
+  });
+
+  if (!response.ok) {
```
elbalexandre, line 83: What is the response status or response body when we try to remove a user that does not exists anymore ?

mbliakharskyi: 204: User will be deleted in the next minutes, depending on how many calls and data are associated to them.
422 error : Server unable to process the request, error will be described in the troubleshoot field.

elbalexandre (approved):

PR #185, opened 2024-03-19
Auth0 Integration

awan7676:


apps/auth0/src/connectors/auth.ts
```diff
@@ -0,0 +1,32 @@
+import { Auth0Error } from './commons/error';
+
+type GetTokenResponseData = {
+  access_token: string;
+  scope: string;
+  expires_in: string;
+  token_type: string;
+};
+
+export const getToken = async (
```
elbalexandre, line 10: From what I am seeing in the documentation, auth0 implement basic oauth with refresh_token.
We should store accessToken + refreshToken and use refreshToken to refresh credentials. Making this will invalidate `get-token` step.

PR #187, opened 2024-03-20
complete jumpcloud

mbliakharskyi:
## Description

## Checklist:


apps/jumpcloud/src/app/install/service.ts
```diff
@@ -0,0 +1,49 @@
...
+  region: string;
+};
+
+export const registerOrganisation = async ({
+  organisationId,
+  apiKey,
+  region,
+}: SetupOrganisationParams) => {
+  const encodedApiKey = await encrypt(apiKey);
+
```
elbalexandre, line 18: We should do a dumb call to check apiKey validity like `await getUsers(...)`

apps/jumpcloud/src/connectors/users.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const response = await fetch(url.toString(), {
+    method: 'DELETE',
+    headers: {
+      'Content-Type': 'application/json',
+      'x-api-key': `${apiKey}`,
+    },
+  });
+
+  if (!response.ok && response.status !== 404 && response.status !== 400) {
```
elbalexandre, line 88: Is it 404 or 400 when a user does not exists ?

apps/jumpcloud/src/inngest/client.ts
```diff
@@ -0,0 +1,37 @@
...
+      data: {
+        organisationId: string;
+        region: string;
+      };
+    };
+    'jumpcloud/users.delete.requested': {
+      data: {
+        userId: string;
+      };
+    };
```
elbalexandre, line 33: We need to pass organisationId (that can be retrieved with `parseWebhookEvent`)
```suggestion
    'jumpcloud/users.delete.requested': {
      data: {
        userId: string;
        organisationId: string
      };
    };
```

apps/jumpcloud/src/inngest/functions/users/delete-users.ts
```diff
@@ -0,0 +1,30 @@
...
+  { id: 'delete-users' },
+  { event: 'jumpcloud/users.delete.requested' },
+  async ({ event }) => {
+    const { userId } = event.data;
+
+    const [organisation] = await db
+      .select({
+        apiKey: Organisation.apiKey,
+      })
+      .from(Organisation);
```
elbalexandre, line 18: I miss that in the previous PR review. But, we should select the org using id retrieved from event data.
```suggestion
      .from(Organisation)
      .where(eq(Organisation.id, organisationId));
```

apps/jumpcloud/src/connectors/users.ts
mbliakharskyi: I tested it.
1. 404 error: invalid user_id
2. 400 error: user not exist.

mbliakharskyi: > I tested it.
> 
> 1. 400 error: invalid user_id
> 2. 404 error: user not exist.

mbliakharskyi: image

mbliakharskyi: image

elbalexandre: Ok so let's drop the 400 error filter, it's ok to throw in this case.

apps/jumpcloud/.env.test
```diff
@@ -0,0 +1,20 @@
+ELBA_SOURCE_ID="0442f541-45d2-487a-9e4b-de03ce4c559e"
+ELBA_API_KEY="abc123dfg456"
+ELBA_API_BASE_URL="http://127.0.0.1:3522/api/rest"
+ELBA_REDIRECT_URL="https://admin.elba.jumplogic/install"
+ELBA_WEBHOOK_SECRET="123"
+
+JUMPCLOUD_API_BASE_URL="https://api-sandbox.jumplogic.app/"
```
elbalexandre, line 7: We should use V2 api if possible

mbliakharskyi: let me check

PR #192, opened 2024-03-25
feat(sharepoint): data protection

tGeorgiy:
## Description
Updated functions / code quality, updated tests, implemented elba.dataProtection.deleteObjects, added cron task to update data protection.

## Checklist:

- [ x ] I have performed a self-review of my own code.
- [ x ] I have commented my code, particularly in hard-to-understand areas.
- [ x ] I have added tests to cover the new feature or fixes.


apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
```diff
@@ -0,0 +1,302 @@
...
+    } else {
+      items.push(item);
+    }
+  });
+
+  return {
+    folders,
+    items,
+  };
+};
```
elbalexandre, line 42: Having items that could items or folders is a bit confusing. Are items items file ? If it's the case let's call them files.

```suggestion
export const groupItems = (
  items: MicrosoftDriveItem[]
) => items.reduce((acc, item) => {
  if (item.file) {
    acc.files.push(item)
  } else {
    acc.folders.push(item)
  }
  return acc
}, { files: [], folders[] });
```

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
```diff
@@ -0,0 +1,302 @@
...
+          email: p.user.email,
+          metadata: {
+            permissionId: permission.id,
+            roles: permission.roles,
+          },
+        });
+      }
+    }
+
+    return formattedPermissions;
```
elbalexandre, line 88:
```suggestion
    return permission.grantedToIdentitiesV2
        .filter(({ user }) => user)
        .map(permissionGrant => ({
          id: `${permission.id}${env.ID_SEPARATOR}${permissionGrant.user.id}`,
          type: 'user',
          displayName: permissionGrant.user.displayName,
          userId: permissionGrant.user.id,
          email: permissionGrant.user.email,
          metadata: {
            permissionId: permission.id,
            roles: permission.roles,
          }
        })
```
Also, I am not sure if we should add the `ID_SEPARATOR` in env variable. If we need to change it, we will have to handle metadata version anyway with code. To do so with env variable could be a lot trickier than without it.

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
```diff
@@ -0,0 +1,302 @@
...
+        return false;
+      });
+
+      if (validPermissions.length) {
+        const dataProtectionItem = {
+          id: item.id,
+          name: item.name,
+          ownerId: item.createdBy.user.id,
+          url: item.webUrl,
+          permissions: validPermissions.map((p) => formatPermissions(p)).flat(),
```
elbalexandre, line 112:
```suggestion
          permissions: validPermissions.map(formatPermissions),

```

`formatPermissions` already returns a flat array, isn't it ?

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
```diff
@@ -0,0 +1,302 @@
...
+): DataProtectionObject[] => {
+  const dataProtection: DataProtectionObject[] = [];
+
+  for (const { item, permissions } of itemsWithPermisions) {
+    if (item.createdBy.user.id) {
+      const validPermissions: MicrosoftDriveItemPermissions[] = permissions.filter((p) => {
+        if (p.link && p.link.scope === 'users') return true;
+        else if (p.grantedToV2?.user) return true;
+        return false;
+      });
```
elbalexandre, line 104:
```suggestion
      const validPermissions: MicrosoftDriveItemPermissions[] = permissions.filter(permission =>
        permission.link?.scope === 'users' || permission.grantedToV2?.user
      });

```

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
```diff
@@ -0,0 +1,302 @@
...
+        });
+      }
+    }
+
+    return formattedPermissions;
+  }
+  return [];
+};
+
+export const parseDataProtetionItems = (
```
elbalexandre, line 93:
```suggestion
export const formatDataProtetionItems = (

```

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
```diff
@@ -0,0 +1,302 @@
...
+        match: 'data.organisationId',
+      },
+    ],
+    retries: env.MICROSOFT_DATA_PROTECTION_SYNC_MAX_RETRY,
+  },
+  { event: 'one-drive/items.sync.triggered' },
+  async ({ event, step }) => {
+    const { siteId, driveId, isFirstSync, folderId, skipToken, organisationId } = event.data;
+
+    logger.info('Sync Items');
```
elbalexandre, line 149:
```suggestion
  async ({ event, step, logger }) => {
    const { siteId, driveId, isFirstSync, folderId, skipToken, organisationId } = event.data;

    logger.info('Sync Items');

```

My bad, I forgot to use the inngest logger in other integration and in the template. If it's used outside a step it will print many times.

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
```diff
@@ -0,0 +1,302 @@
...
+          apiKey: env.ELBA_API_KEY,
+          baseUrl: env.ELBA_API_BASE_URL,
+          region: organisation.region,
+        });
+
+        await elba.dataProtection.updateObjects({
+          objects: dataProtectionItems,
+        });
+      });
+    }
```
elbalexandre, line 262:
```suggestion
    const itemsWithPermisionsResult = await step.run('item-permissions', async () => {
      const itemsWithPermisions: ItemsWithPermisions[] = [];

      const itemsChunks = getCkunkedArray<MicrosoftDriveItem>(
        [...folders, ...items],
        env.MICROSOFT_DATA_PROTECTION_ITEM_PERMISSIONS_CHUNK_SIZE
      );

      for (const itemsChunk of itemsChunks) {
        // eslint-disable-next-line no-await-in-loop -- Avoiding hundreds of inngest functions
        const itemWithPermissionsChunks = await Promise.all(
          itemsChunk.map(async (item) => ({
            ...item,
            permissions: await getItemPermissions({
              token,
              siteId,
              driveId,
              itemId: item.id,
              skipToken: null,
            })
          }))
        );
       itemsWithPermisions.push(...itemWithPermissionsChunks);
      }
      
      const dataProtectionItems = parseDataProtetionItems(
        itemsWithPermisions
      );
      
      if (dataProtectionItems.length === 0) {
        return;
      }
      
      const elba = new Elba({
          organisationId,
          apiKey: env.ELBA_API_KEY,
          baseUrl: env.ELBA_API_BASE_URL,
          region: organisation.region,
        });

        await elba.dataProtection.updateObjects({
          objects: dataProtectionItems,
        });
    });
```
Could be nice to move the chunk fetch logic in a function outside inngest function body.

apps/one-drive/src/connectors/share-point/permissions.ts
```diff
@@ -0,0 +1,148 @@
...
+
+export type GetPermissionsParams = {
+  token: string;
+  siteId: string;
+  driveId: string;
+  itemId: string;
+  skipToken: string | null;
+};
+
+export const getItemPermissions = async ({
```
elbalexandre, line 96: It could be nice to split this function in two:
- getItemPermissions
- getAllItemPermissions

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
tGeorgiy: Items are not exactly files, they can be files and folders, also in Microsoft it called items, but we can call it files, like objects to get permissions and send to elba. Good suggestion with reduce.
Updated this

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
tGeorgiy: `formatPermissions` returns flat array, but then we use map and get array of arrays, so we need to flat this here.
Here we can get one permission or array of permissions from each `(p)` object, so i thought this would be a simplier option.

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
tGeorgiy: Updated this part.

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
tGeorgiy: Updated.

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
tGeorgiy: Updated, thanks for advice.. But I think that we can remove it later, it's more for me to see how function goes

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
tGeorgiy: Updated, but I'm not shure that's what you meant

apps/one-drive/src/connectors/share-point/permissions.ts
tGeorgiy: Updated, but I'm not shure that's what you meant, also need to update tests if it's ok.

tGeorgiy: Also maybe it will be better to do like this:

`  do {
    // eslint-disable-next-line no-await-in-loop -- test
    const { permissions, nextSkipToken: newSkipToken } = await getItemPermissions({
      token,
      siteId,
      driveId,
      itemId,
      skipToken: nextSkipToken,
    });

    allPermissions = [...allPermissions, ...permissions];
    nextSkipToken = newSkipToken;
  } while (nextSkipToken);`
  
  But I need to  disable await-in-loop error

elbalexandre (approved):

tGeorgiy:
ok, for now

apps/one-drive/src/inngest/functions/token/refresh-token.ts
```diff
@@ -5,7 +5,7 @@ import { NonRetriableError } from 'inngest';
 import { db } from '@/database/client';
 import { organisationsTable } from '@/database/schema';
 import { inngest } from '@/inngest/client';
-import { getToken } from '@/connectors/one-drive/auth';
+import { getToken } from '@/connectors/auth/auth';
```
athielba, line 8: ignorable , I am not big fan of keeping the same name for folder and file,  could be 
`import { getToken } from '@/services/authService';` ?

apps/one-drive/src/app/api/webhooks/elba/data-protection/delete-object/route.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const baseSchema = z.object({
+  id: z.string(),
+  organisationId: z.string(),
+  metadata: z.object({
+    siteId: z.string(),
+    driveId: z.string(),
+  }),
+});
+
```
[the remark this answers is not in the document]
tGeorgiy: Will remove delete-object.

apps/one-drive/src/app/api/webhooks/elba/data-protection/delete-object/service.test.ts
```diff
@@ -0,0 +1,56 @@
...
+          siteId,
+          driveId,
+        },
+      },
+    });
+
+    expect(response.status).toBe(200);
+    expect(send).toBeCalledTimes(1);
+    expect(send).toBeCalledWith({
+      name: 'one-drive/data_protection.delete_object.requested',
```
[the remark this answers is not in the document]
tGeorgiy: Nope, I forgot to remove delete-object..

apps/one-drive/src/app/api/webhooks/elba/data-protection/delete-object-permissions/route.ts
```diff
@@ -0,0 +1,29 @@
...
+const baseSchema = z.object({
+  id: z.string(),
+  organisationId: z.string(),
+  metadata: z.object({
+    siteId: z.string(),
+    driveId: z.string(),
+  }),
+  permissions: z.array(z.object({ id: z.string() })),
+});
+
```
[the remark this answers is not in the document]
tGeorgiy: Yes, ok

apps/one-drive/src/connectors/share-point/permissions.ts
```diff
@@ -0,0 +1,163 @@
...
+    const grantedToV2ParseResult = grantedToV2Schema.safeParse(result.data.grantedToV2);
+    const grantedToIdentitiesV2ParseResult = grantedToIdentitiesV2Schema.safeParse(
+      result.data.grantedToIdentitiesV2
+    );
+    if (grantedToV2ParseResult.success) {
+      return {
+        ...result.data,
+        grantedToV2: grantedToV2ParseResult.data,
+      };
+    } else if (grantedToIdentitiesV2ParseResult.success) {
```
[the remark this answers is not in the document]
tGeorgiy: Yes, ok

apps/one-drive/src/inngest/functions/data-protection/delete-item-permission.ts
```diff
@@ -0,0 +1,97 @@
...
+  { event: 'one-drive/data_protection.delete_object_permissions.requested' },
+  async ({ event, step, logger }) => {
+    const {
+      id: itemId,
+      organisationId,
+      metadata: { siteId, driveId },
+      permissions,
+    } = event.data;
+
+    logger.info('Delete Start');
```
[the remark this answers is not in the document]
tGeorgiy: This is only for my debug, to watch how process goes, I wanted to remove it later, but can remove now.

apps/one-drive/src/inngest/functions/data-protection/delete-item-permission.ts
```diff
@@ -0,0 +1,97 @@
...
+          }
+          /* eslint-enable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-member-access -- End of error handling */
+        })
+      )
+    );
+
+    return permissionDeletionResults.reduce<{
+      deletedPermissions: string[];
+      notFoundPermissions: string[];
+      unexpectedFailedPermissions: string[];
```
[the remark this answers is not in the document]
tGeorgiy: As I understand, it should do a retry in step.run before goes to return, this array `unexpectedFailedPermissions` is needed more for tracking in the inngest. But I can remove it.

apps/one-drive/src/inngest/functions/token/refresh-token.ts
tGeorgiy: I can rename auth.ts to get-token.ts for example.
I think that this parts like auth and users will be the separate module, because it's the same in few integrations;

apps/one-drive/src/connectors/share-point/sites.test.ts
```diff
@@ -0,0 +1,78 @@
...
+const sites: MicrosoftSite[] = Array.from({ length: env.SITES_SYNC_BATCH_SIZE }, (_, i) => ({
+  id: `site-id-${i}`,
+}));
+
+describe('sites connector', () => {
+  describe('getSites', () => {
+    // mock token API endpoint using msw
+    beforeEach(() => {
+      server.use(
+        http.get(`${env.MICROSOFT_API_URL}/sites`, ({ request }) => {
```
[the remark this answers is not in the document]
tGeorgiy: I don't know for sure, because I have different links, different params in most cases and different responses in few tests.

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
```diff
@@ -0,0 +1,361 @@
+import type { DataProtectionObject, DataProtectionObjectPermission } from '@elba-security/sdk';
```
athielba, line 1: this event seems complex and, IMO  code can be split in  small & and separate

apps/one-drive/src/middleware.ts
```diff
@@ -0,0 +1,17 @@
+import { NextResponse } from 'next/server';
```
athielba, line 1: need to be updated

apps/one-drive/src/inngest/middlewares/rate-limit-middleware.ts
```diff
@@ -0,0 +1,40 @@
+import { InngestMiddleware, RetryAfterError } from 'inngest';
```
athielba, line 1: check zoom please

apps/one-drive/src/inngest/functions/users/sync-users.ts
```diff
@@ -0,0 +1,104 @@
...
+        logger.warn('Retrieved users contains invalid data', {
+          organisationId,
+          tenantId: organisation.tenantId,
+          invalidUsers: result.invalidUsers,
+        });
+      }
+
+      await elba.users.update({
+        users: result.validUsers.map(formatElbaUser),
+      });
```
athielba, line 79:
```suggestion
     if(result.validUsers.length > 0) {
      await elba.users.update({
        users: result.validUsers.map(formatElbaUser),
      });
     
     }
     
```

apps/one-drive/src/inngest/functions/token/refresh-token.ts
```diff
@@ -0,0 +1,69 @@
...
+        match: 'data.organisationId',
+      },
+    ],
+    retries: env.TOKEN_REFRESH_MAX_RETRY,
+  },
+  { event: 'one-drive/token.refresh.requested' },
+  async ({ event, step }) => {
+    const { organisationId, expiresAt } = event.data;
+
+    await step.sleepUntil('wait-before-expiration', subMinutes(new Date(expiresAt), 5));
```
athielba, line 35: it is better to refresh the toke 15 min early, because if is stayed in the queue more than 5, it would have been already expired.  You can set it based on it's life time

apps/one-drive/src/inngest/functions/token/refresh-token.ts
```diff
@@ -0,0 +1,69 @@
...
+          tenantId: organisationsTable.tenantId,
+        })
+        .from(organisationsTable)
+        .where(and(eq(organisationsTable.id, organisationId)));
+
+      if (!organisation) {
+        throw new NonRetriableError(`Could not retrieve organisation with id=${organisationId}`);
+      }
+
+      const { token, expiresIn } = await getToken(organisation.tenantId);
```
athielba, line 49: How does this work without refresh token ? according to this [Doc](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow#successful-response-2) I could see the refresh_token ? but I don't see that you are passing the refresh token  ? Does it support different way ??

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
```diff
@@ -0,0 +1,361 @@
...
+        const dataProtectionItem = {
+          id: item.id,
+          name: item.name,
+          ownerId: item.createdBy.user.id,
+          url: item.webUrl,
+          metadata: {
+            siteId,
+            driveId,
+          },
+          permissions: validPermissions.map(formatPermissions).flat(),
```
athielba, line 163: I think we can sent the following info if the API does support
```
export const updateDataProtectionObjectsSchema = z.object({
  objects: z.array(
    z.object({
      id: z.string().min(1),
      name: z.string().min(1),
      lastAccessedAt: z.string().datetime().optional(),
      url: z.string().url(),
      ownerId: z.string().min(1),
      metadata: jsonSchema,
      contentHash: z.string().min(1).optional(),
      isSensitive: z.boolean().optional(),
      updatedAt: z.string().datetime().optional(),
      permissions: z.array(permissionSchema),
    })
  ),
});
```

[Updated at ](https://learn.microsoft.com/en-us/onedrive/developer/rest-api/api/driveitem_get?view=odsp-graph-online#response-1)

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
```diff
@@ -0,0 +1,361 @@
...
+  return chunks;
+};
+
+export const formatPermissions = (
+  permission: MicrosoftDriveItemPermissions
+): DataProtectionObjectPermission[] | [] => {
+  if (permission.grantedToV2?.user) {
+    return [
+      {
+        id: permission.id,
```
athielba, line 65: We have three types 3 types of permissions ? you should set the permissions based on this details
https://github.com/elba-security/elba-security/blob/tGeorgiy/onedrive/docs/api/data-protection/update-objects.md#type-of-permissions, for example the `user` role  will support the following details, not everything together 
```json
{
  "id": "permission-id-2",
  "type": "user",
  "email": "user-email-id@alpha.com",
}

or

{
  "id": "permission-id-2",
  "type": "user",
  "userId": "user-id",
  "displayName": "display-name-of-the-user",
}
```

apps/one-drive/scripts/run-test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
tGeorgiy: updated

apps/one-drive/src/app/api/inngest/route.ts
```diff
@@ -0,0 +1,15 @@
+import { serve } from 'inngest/next';
+import { env } from '@/env';
+import { inngest } from '@/inngest/client';
+import { inngestFunctions } from '@/inngest/functions';
+
+export const preferredRegion = env.VERCEL_PREFERRED_REGION;
```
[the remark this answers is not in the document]
tGeorgiy: updated

apps/one-drive/src/app/api/webhooks/elba/data-protection/delete-object-permissions/route.ts
```diff
@@ -0,0 +1,32 @@
+import { NextResponse } from 'next/server';
+import { z } from 'zod';
+import { parseWebhookEventData } from '@elba-security/sdk';
+import { env } from '@/env';
+import { deleteObjectPermissions } from './service';
+
+export const preferredRegion = env.VERCEL_PREFERRED_REGION;
```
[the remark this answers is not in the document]
tGeorgiy: updated

apps/one-drive/src/app/api/webhooks/microsoft/event-handler/route.ts
```diff
@@ -0,0 +1,44 @@
...
+  if (req.nextUrl.searchParams.get('validationToken')) {
+    return new NextResponse(req.nextUrl.searchParams.get('validationToken'), {
+      status: 200,
+      headers: {
+        'Content-Type': 'text/plain',
+      },
+    });
+  }
+
+  const data = (await req.json()) as WebhookResponse<SubscriptionPayload>;
```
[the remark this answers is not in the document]
tGeorgiy: yes, we can use unknown

apps/one-drive/src/app/auth/route.ts
```diff
@@ -0,0 +1,59 @@
+import type { NextRequest } from 'next/server';
+import { z } from 'zod';
+import { logger } from '@elba-security/logger';
+import { ElbaInstallRedirectResponse } from '@elba-security/nextjs';
+import { env } from '@/env';
+import { setupOrganisation } from './service';
+
+export const preferredRegion = env.VERCEL_PREFERRED_REGION;
```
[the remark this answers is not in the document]
tGeorgiy: updated

apps/one-drive/src/app/install/route.ts
```diff
@@ -0,0 +1,37 @@
...
+      region: request.nextUrl.searchParams.get('region'),
+    });
+
+    cookies().set('organisation_id', organisationId);
+    cookies().set('region', region);
+  } catch (error) {
+    logger.warn('Could not redirect user to Microsoft app install url', {
+      error,
+    });
+    redirect(`${env.ELBA_REDIRECT_URL}?source_id=${env.ELBA_SOURCE_ID}&error=internal_error`);
```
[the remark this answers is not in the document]
tGeorgiy: Updated

apps/one-drive/src/common/crypto.ts
```diff
@@ -0,0 +1,16 @@
...
+import { env } from '@/env';
+
+export const encrypt = (data: string) => {
+  return encryptText({ data, key: env.ENCRYPTION_KEY });
+};
+
+export const decrypt = (data: string) => {
+  try {
+    return decryptText({ data, key: env.ENCRYPTION_KEY });
+  } catch (error) {
```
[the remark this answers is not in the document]
tGeorgiy: Updated

apps/one-drive/src/database/client.ts
```diff
@@ -0,0 +1,19 @@
+import { Pool, neonConfig } from '@neondatabase/serverless';
```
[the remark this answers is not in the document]
tGeorgiy: Updated

apps/one-drive/src/env.ts
```diff
@@ -0,0 +1,76 @@
...
+    MICROSOFT_CLIENT_ID: z.string().min(1),
+    MICROSOFT_CLIENT_SECRET: z.string().min(1),
+    MICROSOFT_REDIRECT_URI: z.string().url(),
+    MICROSOFT_INSTALL_URL: z
+      .string()
+      .url()
+      .default('https://login.microsoftonline.com/organizations/adminconsent'),
+    MICROSOFT_API_URL: z.string().url().default('https://graph.microsoft.com/v1.0'),
+    MICROSOFT_AUTH_API_URL: z.string().url().default('https://login.microsoftonline.com'),
+    MICROSOFT_DATA_PROTECTION_ITEMS_SYNC_CONCURRENCY: z.coerce
```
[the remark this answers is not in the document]
tGeorgiy: Updated

apps/one-drive/src/inngest/functions/data-protection/delete-item-permission.ts
```diff
@@ -0,0 +1,90 @@
...
+import { db } from '@/database/client';
+import { organisationsTable } from '@/database/schema';
+import { env } from '@/env';
+import { inngest } from '@/inngest/client';
+import { decrypt } from '@/common/crypto';
+import { deleteItemPermission } from '@/connectors/share-point/delete-item-permission';
+
+export const deleteDataProtectionItemPermissions = inngest.createFunction(
+  {
+    id: 'delete-data-protection-object-permissions',
```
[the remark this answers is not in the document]
tGeorgiy: Updated

apps/one-drive/src/inngest/functions/data-protection/refresh-item.ts
```diff
@@ -0,0 +1,96 @@
...
+import { inngest } from '@/inngest/client';
+import { decrypt } from '@/common/crypto';
+import { getAllItemPermissions } from '@/connectors/share-point/permissions';
+import { getItem } from '@/connectors/share-point/item';
+import { getElbaClient } from '@/connectors/elba/client';
+import { formatDataProtetionItems } from './sync-items';
+
+export const refreshItem = inngest.createFunction(
+  {
+    id: 'refresh-data-protection-objects',
```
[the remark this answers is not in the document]
tGeorgiy: Updated

apps/one-drive/src/middleware.ts
tGeorgiy: Updated

apps/one-drive/src/inngest/functions/token/refresh-token.ts
tGeorgiy: It's Alex implementation, of microsoft auth, we are getting new token each time, without refresh. He told that it`s ok.

apps/one-drive/src/common/pagination.ts
```diff
@@ -0,0 +1,16 @@
+import { z } from 'zod';
```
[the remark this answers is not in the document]
tGeorgiy: updated

apps/one-drive/src/connectors/auth/get-token.test.ts
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable @typescript-eslint/no-unsafe-call -- test conveniency */
```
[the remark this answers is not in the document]
tGeorgiy: updated

apps/one-drive/src/inngest/functions/token/refresh-token.ts
tGeorgiy: updated

apps/one-drive/src/connectors/delta/get-delta.ts
```diff
@@ -0,0 +1,59 @@
...
+    headers: {
+      Authorization: `Bearer ${token}`,
+    },
+  });
+
+  if (!response.ok) {
+    throw new MicrosoftError('Could not retrieve delta', { response });
+  }
+
+  const data = (await response.json()) as MicrosoftPaginatedResponse<Delta>;
```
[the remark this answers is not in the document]
tGeorgiy: Updated this part

apps/one-drive/src/inngest/functions/data-protection/delete-item-permission.ts
```diff
@@ -0,0 +1,90 @@
...
+    concurrency: {
+      key: 'event.data.organisationId',
+      limit: 1,
```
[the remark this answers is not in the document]
tGeorgiy: updated this part

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
```diff
@@ -0,0 +1,361 @@
...
+    });
+
+    if (nextSkipToken) {
+      await step.sendEvent('sync-next-items-page', {
+        name: 'one-drive/items.sync.triggered',
+        data: {
+          ...event.data,
+          folder: {
+            id: folder?.id || null,
+            paginated: folder?.paginated || getPermissionsResult.setPaginated,
```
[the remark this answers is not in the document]
tGeorgiy: I might need to work a bit on the naming, but the logic here is a bit complex. These are the permissions for this folder, which are needed to filter inherited permissions, and it can be empty array, so i need to check if it is a paginated call I don't need to get permissions one more time. I will think about how this can be simplified.

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
tGeorgiy: Do you mean that I should split the code into smaller functions, or separate the logic and possibly add some additional inngest functions?

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
tGeorgiy: ok, tnx, will update this.

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
athielba: OK, that is fine, just I wanted to point  the `updatedAt`

apps/one-drive/src/inngest/middlewares/rate-limit-middleware.ts
tGeorgiy: I'm not shure that I understand what I need to do.
The main difference is that in zoom we check that if error is instance of ZoomError then we do nothing, but if it`s not we need to check status and if needed to set RetryAfterError.
So I need to implement this approach right?

athielba: retryafter should be like below `${retryafter}s`, it should be same as zoom

tGeorgiy: ok, I saw that, and want to do like in zoom, but not quite shure that it was all what I need.

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
tGeorgiy: updated, splitted in different files

apps/one-drive/src/inngest/middlewares/rate-limit-middleware.ts
tGeorgiy: updated

apps/one-drive/src/inngest/functions/data-protection/sync-items.ts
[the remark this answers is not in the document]
tGeorgiy: little refactor and renaming was done.

apps/one-drive/src/inngest/client.ts
```diff
@@ -0,0 +1,150 @@
...
+      data: {
+        siteId: string;
+        driveId: string;
+        subscriptionId: string;
+        tenantId: string;
+        skipToken: string | null;
+      };
+    };
+  }>(),
+  middleware: [rateLimitMiddleware, unauthorizedMiddleware, sentryMiddleware],
```
elbalexandre, line 148:
```suggestion
  middleware: [rateLimitMiddleware, sentryMiddleware],
```
We should unplug `unauthorizedMiddleware` for now as Microsoft can respond with a 401 randomly. We will implement a better way to handle it later.

apps/one-drive/src/connectors/one-drive/share-point/permissions.ts
```diff
@@ -0,0 +1,171 @@
...
+  | (Omit<z.infer<typeof basePSchema>, 'grantedToV2'> & {
+      grantedToV2: z.infer<typeof grantedToV2Schema>;
+    })
+  | (Omit<z.infer<typeof basePSchema>, 'grantedToIdentitiesV2'> & {
+      grantedToIdentitiesV2: z.infer<typeof grantedToIdentitiesV2Schema>;
+    })
+  | null => {
+  const result = basePSchema.safeParse(data);
+
+  if (result.success) {
```
elbalexandre, line 70: *Skippable* that is just a tip to make your code more readable.
You can early returns in this kind of situation:

```
if (!result.success) {
  return null
}

const grantedToV2ParseResult = // ...
// ...
```

apps/one-drive/src/connectors/one-drive/share-point/item.ts
```diff
@@ -0,0 +1,38 @@
...
+    },
+  });
+
+  if (!response.ok) {
+    if (response.status === 404) {
+      return 'notFound';
+    }
+
+    throw new MicrosoftError('Could not retrieve item', { response });
+  }
```
elbalexandre, line 33:
```suggestion
  if (response.status === 404) {
    return 'notFound';
  }
    
  if (!response.ok) {
    throw new MicrosoftError('Could not retrieve item', { response });
  }
```

Response.ok is false when status is not a 2XX only, so it's okay to check for 404 before checking if the response is ok.

Skippable: Regarding returning 'notFound', I personally found it cleaner to returns just `null`.

apps/one-drive/src/connectors/one-drive/share-point/delete-item-permission.ts
```diff
@@ -0,0 +1,33 @@
...
+
+type DeleteItemPermissionParams = {
+  itemId: string;
+  token: string;
+  siteId: string;
+  driveId: string;
+  permissionId: string;
+};
+
+export const deleteItemPermission = async ({
```
elbalexandre, line 12: I think this could have been moved into `permissions.ts`.

apps/one-drive/src/connectors/one-drive/share-point/delete-item-permission.ts
```diff
@@ -0,0 +1,33 @@
...
+  );
+
+  const response = await fetch(url, {
+    method: 'DELETE',
+    headers: {
+      Authorization: `Bearer ${token}`,
+    },
+  });
+
+  if (!response.ok) {
```
elbalexandre, line 30: Do we really want to throw an error when the permission is not found ? Perhaps the user deleted it by itself before this function is called.

apps/one-drive/src/inngest/functions/data-protection/delete-item-permission.ts
```diff
@@ -0,0 +1,90 @@
...
+              token,
+              siteId,
+              driveId,
+              itemId,
+              permissionId,
+            });
+
+            return { status: 204, permissionId };
+            /* eslint-disable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-member-access -- Start of error handling */
+          } catch (error: any) {
```
elbalexandre, line 65:
```suggestion
          } catch (error) {
          if (error instanceof MicrosoftError && error.response.status === 404) {
          
          }
```

apps/one-drive/src/inngest/functions/data-protection/delete-item-permission.ts
```diff
@@ -0,0 +1,90 @@
+import { eq } from 'drizzle-orm';
+import { NonRetriableError } from 'inngest';
+import { db } from '@/database/client';
+import { organisationsTable } from '@/database/schema';
+import { inngest } from '@/inngest/client';
+import { decrypt } from '@/common/crypto';
+import { deleteItemPermission } from '@/connectors/one-drive/share-point/delete-item-permission';
+import { env } from '@/common/env';
+
```
elbalexandre, line 9: The name of the file is in singular, but we are deleting multiple permissions.

apps/one-drive/src/inngest/functions/data-protection/delete-item-permission.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (!organisation) {
+      throw new NonRetriableError(`Could not retrieve organisation with itemId=${organisationId}`);
+    }
+
+    const token = await decrypt(organisation.token);
+
+    const permissionDeletionResults = await Promise.allSettled(
+      permissions.map((permissionId) =>
+        step.run('delete-item-permission', async () => {
```
elbalexandre, line 53: How many permissions are we deleting at ounce ? Like max amount we can have ?

loop with step.run could be a bit slow.

@proquig What do you think of using another function that remove a single permissions and use `step.invoke` ? Is it worth it ?

apps/one-drive/src/inngest/functions/data-protection/delete-item-permission.ts
```diff
@@ -0,0 +1,90 @@
...
+    }>(
+      (acc, el) => {
+        if (el.status === 'fulfilled') {
+          if (el.value.status === 204) acc.deletedPermissions.push(el.value.permissionId);
+          if (el.value.status === 404) acc.notFoundPermissions.push(el.value.permissionId);
+        }
+
+        return acc;
+      },
+      { deletedPermissions: [], notFoundPermissions: [] }
```
elbalexandre, line 87: I guess having the list of permissions deletion that failed for other reason than 404 could be a must have to debug.

apps/one-drive/src/inngest/functions/data-protection/refresh-item.ts
```diff
@@ -0,0 +1,96 @@
...
+          {
+            item,
+            permissions,
+          },
+        ],
+        siteId,
+        driveId,
+      });
+
+      if (!dataProtectionItem.length) return;
```
elbalexandre, line 85: We should delete the object in this case as well.

apps/one-drive/src/inngest/functions/data-protection/refresh-item.ts
```diff
@@ -0,0 +1,96 @@
...
+
+      await elba.dataProtection.updateObjects({
+        objects: dataProtectionItem,
+      });
+    });
+
+    return {
+      status: 'completed',
+    };
```
elbalexandre, line 94: We don't really need to returns a status here:  it will always be completed and inngest function run already has a status (scheduled | running | completed). The status returned in other functions was for pagination purpose.
```suggestion
```

apps/one-drive/src/inngest/functions/data-protection/update-items.ts
```diff
@@ -0,0 +1,161 @@
...
+              itemWithPermisions.item.name !== 'root' &&
+              !toDelete.includes(itemWithPermisions.item.id)
+            )
+              acc.push(itemWithPermisions.item.id);
+            return acc;
+          }, []);
+
+          reduced.push(...toDelete);
+
+          return reduced;
```
elbalexandre, line 108:
```suggestion
          return itemsWithPermisions.reduce<string[]>((acc, itemWithPermisions) => {
            if (
              !itemWithPermisions.permissions.length &&
              itemWithPermisions.item.name !== 'root' &&
              !toDelete.includes(itemWithPermisions.item.id)
            )
              acc.push(itemWithPermisions.item.id);
            return acc;
          }, [...toDelete]);
```

or 

```suggestion
          return [
            ...toDelete,
            ...itemsWithPermisions
              .filter(() =>/* filter permissions*/)
              .map(({ item }) => item.id);
          ]
```

apps/one-drive/src/inngest/functions/data-protection/update-items.ts
```diff
@@ -0,0 +1,161 @@
...
+          skipToken: nextSkipToken,
+        },
+      });
+
+      return {
+        status: 'ongoing',
+      };
+    }
+
+    if (!newDeltaToken) throw new MicrosoftError('Delta token not found!');
```
elbalexandre, line 141: We are we throwing this late ?
We got the deltaToken sooner ?

I think we should throw `NonRetrieableError` instead of MicrosoftError.

apps/one-drive/src/inngest/functions/delta/initialize-delta.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    await db
+      .insert(sharePointTable)
+      .values({
+        organisationId,
+        siteId,
+        driveId,
+        subscriptionId: data.id,
+        subscriptionExpirationDate: data.expirationDateTime,
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- cant be null
```
elbalexandre, line 96: Are you sure about this ?
Perhaps we should throw sooner if newDeltaToken is falsy ?

elbalexandre (commented):
We should not forget to rename this integration (np for the branch name) and squash migration.

@yuriiyf Please, could you review the part on subscription ?

yuriiyf (commented):
Looks good, the only question is about checking that the request  to webhook (`api/microsoft...`) is made by MS only. Not sure about your discussions about that

apps/one-drive/src/inngest/functions/data-protection/common/helpers.ts
```diff
@@ -0,0 +1,244 @@
...
+      (permission) => !parentPermissionIds.includes(permission.id)
+    );
+
+    acc.push({
+      item: itemWithPermisions.item,
+      permissions: filteredPermissions,
+    });
+
+    return acc;
+  }, []);
```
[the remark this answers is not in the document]
tGeorgiy: good suggestion, updated

apps/one-drive/src/inngest/functions/data-protection/common/helpers.ts
```diff
@@ -0,0 +1,244 @@
...
+
+  if (folder?.id && !folder.paginated) {
+    const { permissions } = await getAllItemPermissions({
+      token,
+      siteId,
+      driveId,
+      itemId: folder.id,
+    });
+
+    global.repeats++;
```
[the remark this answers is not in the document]
tGeorgiy: forgot to remove this

apps/one-drive/src/inngest/functions/data-protection/common/helpers.ts
```diff
@@ -0,0 +1,244 @@
...
+    parentFolderPermissions.push(...permissions.map((el) => el.id));
+  } else if (folder?.permissions.length) {
+    parentFolderPermissions.push(...folder.permissions);
+  }
+
+  return {
+    parentFolderPermissions,
+    parentFolderPaginated,
+  };
+};
```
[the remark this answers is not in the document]
tGeorgiy: good suggestion, updated

apps/one-drive/src/inngest/client.ts
```diff
@@ -0,0 +1,150 @@
...
+    'one-drive/app.install.requested': {
+      data: {
+        organisationId: string;
+      };
+    };
+    'one-drive/app.uninstalled.requested': {
+      data: {
+        organisationId: string;
+      };
+    };
```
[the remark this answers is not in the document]
tGeorgiy: updated this)

apps/one-drive/src/connectors/one-drive/share-point/permissions.ts
tGeorgiy: good suggestion, updated

apps/one-drive/src/connectors/one-drive/share-point/item.ts
tGeorgiy: good suggestion, updated

apps/one-drive/src/connectors/one-drive/share-point/delete-item-permission.ts
tGeorgiy: updated, but maybe we need to do it in this way with subscriptions also?

apps/one-drive/src/connectors/one-drive/share-point/delete-item-permission.ts
tGeorgiy: In this case response will have status 204, it doesn't matter if there was permission on the file or not.
It will try to remove permission from the file, if there was permission or not, it doesn't matter.
So it will fail if file was deleted or token is invalid. I'm not shoure, maybe we need to throw NonRetriableError ?

apps/one-drive/src/inngest/functions/data-protection/delete-item-permission.ts
tGeorgiy: updated

apps/one-drive/src/inngest/functions/data-protection/delete-item-permission.ts
tGeorgiy: updated

apps/one-drive/src/inngest/functions/data-protection/delete-item-permission.ts
tGeorgiy: Implemented)

apps/one-drive/src/inngest/functions/data-protection/refresh-item.ts
tGeorgiy: good suggestion, updated

apps/one-drive/src/inngest/functions/data-protection/update-items.ts
tGeorgiy: good suggestion, I like first variant) updated

apps/one-drive/src/inngest/functions/data-protection/update-items.ts
tGeorgiy: Updated to `NonRetrieableError`.
The main problem why I can't throw error in the connector is that the deltaToken appears only on last pagination page.
So I should do all the stuff from previous pages and I don't know if I get the deltaToken, it should be there in all cases, but throwing error here prevents updating db record with faulty values (and db error).

So if I have no `skipToken`, I should have `deltaToken` then.

apps/one-drive/src/inngest/functions/delta/initialize-delta.ts
tGeorgiy: Added  `if (!newDeltaToken) throw new NonRetriableError('Delta token not found!');` before `step.invoke('one-drive/drives.subscription.triggered'`

apps/one-drive/src/inngest/functions/data-protection/update-items.ts
elbalexandre: Could you add a comment on top of the function or in the function so this knownledge is not lost ?

apps/one-drive/src/connectors/one-drive/share-point/delete-item-permission.ts
elbalexandre: It's a picky comment, I guess so. I am not a fan of huge file, but I am not a fan of unnecessary splitting as well.
Your call on that.

apps/one-drive/src/connectors/one-drive/share-point/delete-item-permission.ts
elbalexandre: I think if the file is not found we should delete it using `elba.dataProtection.deleteObject({ ids: [fileId] })`.

apps/one-drive/src/inngest/functions/data-protection/delete-item-permission.ts
elbalexandre: Ok, let keep this comment unresolve so Guillaume has a chance to look at it.

apps/one-drive/src/connectors/one-drive/share-point/delete-item-permission.ts
tGeorgiy: i think I can move it all to one file, there are small functions, so we don't need to split if we want smaller amount of files. will update this.

apps/one-drive/src/connectors/one-drive/share-point/delete-item-permission.ts
tGeorgiy: ok, I will update this, if it got 404, will remove it from Elba

apps/one-drive/src/inngest/functions/data-protection/update-items.ts
tGeorgiy: Yes, ok!

proquig (approved):

PR #199, opened 2024-03-28
feat: google integration

proquig:
## Description

This contains Google integration

It includes:
- authentication
- data protection
- third party apps

The edge runtime is not used as the Google libraries are not compatible

## Additional Notes

N/A

## Checklist:


apps/google/src/app/auth/service.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    customerId = user.customerId;
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any -- error handling
+  } catch (error: any) {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access -- error handling
+    if (error?.errors?.[0]?.reason === 'forbidden') {
+      throw new Error('User is not admin');
+    }
+
```
elbalexandre, line 44: RequestError can be imported from `google-auth-library` (https://github.com/googleapis/google-auth-library-nodejs/blob/main/src/transporters.ts)
```suggestion
  } catch (error) {
    if (error instanceof RequestError && error.errors.[0]?.reason === 'forbidden') {
      throw new Error('User is not admin');
    }

```

apps/google/src/app/auth/service.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (!user.isAdmin) {
+      throw new Error('User is not admin');
+    }
+
+    if (!user.customerId) {
+      throw new Error('Missing Google customer id');
+    }
+
+    customerId = user.customerId;
```
elbalexandre, line 36: optional (no let)
```suggestion
    return { customerId: user.customerId, email };
```

apps/google/src/connectors/google/members.ts
```diff
@@ -0,0 +1,54 @@
...
+  return { members, nextPageToken };
+};
+
+export const listAllGoogleMembers = async (
+  listMembersParams: adminDirectory.Params$Resource$Members$List
+) => {
+  let pageToken: string | undefined;
+  const members: GoogleMember[] = [];
+  do {
+    // eslint-disable-next-line no-await-in-loop -- we loop through every permissions
```
elbalexandre, line 44:
```suggestion
    // eslint-disable-next-line no-await-in-loop -- we loop through every organisation members
```

apps/google/src/connectors/google/users.ts
```diff
@@ -0,0 +1,89 @@
...
+    query,
+  });
+
+  const users: T[] = [];
+  for (const user of googleUsers || []) {
+    const result = schema.safeParse(user);
+    if (result.success) {
+      users.push(result.data);
+    }
+  }
```
elbalexandre, line 73: (not even a change request)
We are extensively using this pattern in many integration.
Perhaps we could do something suggested [here](https://github.com/colinhacks/zod/discussions/672)

apps/google/src/inngest/functions/data-protection/sync-shared-drives.ts
```diff
@@ -0,0 +1,203 @@
...
+
+          if (manager.type !== 'group') {
+            return { sharedDriveId, managerUserId: null, managerEmail: manager.emailAddress };
+          }
+
+          const [groupOwner] = await listAllGoogleMembers({
+            auth: authClient,
+            groupKey: manager.emailAddress,
+            roles: 'OWNER',
+            maxResults: 200, // Set maxResults to max value
```
elbalexandre, line 94: Why 200 here, if we use only the first entry returned ?

apps/google/src/inngest/functions/data-protection/refresh-object.ts
```diff
@@ -0,0 +1,116 @@
...
+      const object = formatDataProtectionObject({ file, owner: ownerId, permissions });
+
+      await elba.dataProtection.updateObjects({ objects: [object] });
+
+      return { status: 'updated' };
+
+      /* eslint-disable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-member-access -- Start of error handling */
+    } catch (error: any) {
+      if (error?.code === 404) {
+        /* eslint-enable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-member-access -- End of error handling */
```
elbalexandre, line 106: GAxiosError comes from `Common`, the other class specified earlier can also be used (it's extending from GAxiosError)
```suggestion
    } catch (error) {
      if (error instanceof GAxiosError && error.code === 404) {
```

apps/google/src/inngest/functions/data-protection/refresh-object.ts
```diff
@@ -0,0 +1,116 @@
...
+          // eslint-disable-next-line @typescript-eslint/no-loop-func -- we iterate through pageToken
+          async () => {
+            return listGoogleFileNonInheritedPermissions({
+              auth: authClient,
+              fileId: objectId,
+              pageToken,
+              pageSize: 100,
+            });
+          }
+        );
```
elbalexandre, line 91: Do we really have file with that much permissions ?
Perhaps we could have just do a connector that retrieve all file non inherited permissions ?

apps/google/src/inngest/functions/data-protection/refresh-object.ts
```diff
@@ -0,0 +1,116 @@
...
+      await elba.dataProtection.updateObjects({ objects: [object] });
+
+      return { status: 'updated' };
+
+      /* eslint-disable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-member-access -- Start of error handling */
+    } catch (error: any) {
+      if (error?.code === 404) {
+        /* eslint-enable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-member-access -- End of error handling */
+
+        await elba.dataProtection.deleteObjects({ ids: [objectId] });
```
elbalexandre, line 108: Not really an opinionated comment, but I returning null in the connector when an entities does not exists anymore. It's handy as it allows to not handle the SaaS related stuff in the inngest function and it make its code easier to read.

apps/google/src/inngest/functions/data-protection/delete-object-permissions.ts
```diff
@@ -0,0 +1,130 @@
...
+              { code: 403, reason: 'insufficientFilePermissions' },
+            ] as const;
+            for (const ignoredError of ignoredErrors) {
+              if (
+                error?.code === ignoredError.code &&
+                error?.errors?.[0]?.reason === ignoredError.reason
+              ) {
+                return { status: 'ignored', reason: ignoredError.reason };
+              }
+            }
```
elbalexandre, line 87: Could be better for readability to handle thoses in the connector

apps/google/src/app/dwd/service.ts
```diff
@@ -0,0 +1,47 @@
...
+    });
+
+  await inngest.send({
+    name: 'google/users.sync.requested',
+    data: {
+      isFirstSync: true,
+      organisationId,
+      syncStartedAt: new Date().toISOString(),
+      pageToken: null,
+    },
```
elbalexandre, line 43: There is no install event to cancel running function for the org. What are the impact of a double installation here ?

elbalexandre (approved):

apps/google/src/inngest/functions/data-protection/sync-shared-drives.ts
proquig: Because as we can't filter out non-users and inactive ones directly in the request, it happens after we get the list where we make sure with zod that we only return active users.

apps/google/src/inngest/functions/data-protection/refresh-object.ts
proquig: Yeah permissions list can be huge sometimes. I think I forgot to update the code after I implemented the function to retrieve every permission.

apps/google/src/app/auth/service.ts
proquig: If you look at the index it's not exported. I tried to check for GaxiosError as well, a solution could be to put the common library in peer dependencies but annoying

n2ctech (approved):

PR #210, opened 2024-04-05
chore: set config path for middleware to avoid calling middleware for every route

proquig:
## Description

This improves the middleware to only call it when the path is matching the config

## Additional Notes

N/A

## Checklist:


packages/nextjs/src/middleware.ts
```diff
@@ -7,15 +7,16 @@ export type CreateElbaMiddlewareOptions = {
...
-      try {
-        await validateWebhookRequestSignature(request, webhookSecret);
-      } catch (error) {
-        logger.error('Could not validate webhook request signature', { error, request });
-        return new NextResponse(null, { status: 401, statusText: 'unauthorized' });
-      }
+export const createElbaMiddleware = ({
+  webhookSecret,
+}: CreateElbaMiddlewareOptions): { config: { matcher: string[] }; middleware: NextMiddleware } => ({
+  config: { matcher: ['/api/webhook/elba/(.*)', '/api/webhooks/elba/(.*)'] },
```
proquig, line 13: Sadly nextjs is inconvenient while parsing the middleware config, the config needs to be exported in the middleware file, importing it from another file like here is not supported for example

proquig: Had no choice, I had to specify the config in every middleware to make it work properly
image

elbalexandre (approved):

PR #212, opened 2024-04-09
chore: improve setup

proquig:
## Description

This removes some common scripts and configurations which were duplicated between integrations
This adds `testcontainers` to run the docker compose stack before starting tests. As a consequence we don't need to configure it with environment variables and there won't be ports conflicts while running the stacks simultaneously.

## Additional Notes

N/A

## Checklist:


apps/slack/vitest.config.mts
```diff
@@ -1,20 +1,26 @@
...
+
 export default defineConfig({
   test: {
-    setupFiles: ['./vitest/setup-database.ts', './vitest/setup-msw-handlers.ts'],
-    env: process.env,
-    // use 'node' if your integration is not compatible with edge runtime
-    environment: 'edge-runtime',
+    globalSetup: '@elba-security/test-utils/vitest/global-setup',
+    setupFiles: [
+      './vitest/setup-database.ts',
```
athielba, line 20: I am wondering, why it is different from other saas setup (why not from packages) ?

packages/test-utils/src/vitest/global-setup.ts
```diff
@@ -0,0 +1,42 @@
...
+  const db = drizzle(sql);
+
+  await migrate(db, { migrationsFolder: './drizzle' }).finally(void sql.end);
+};
+
+export const teardown = async () => {
+  if (stack) {
+    await stack.down();
+  }
+};
```
athielba, line 42: Nice 👏🏼 👏🏼

drizzle.config.ts
```diff
@@ -0,0 +1,13 @@
+import type { Config } from 'drizzle-kit';
+import { config } from 'dotenv';
+
+config({ path: '.env.local' });
+
```
athielba, line 5: if would be nice if you could check this here too
```
const { error } = config({ path: '.env.local' });

if (error) {
  throw new Error(`Could not find environment variables file: .env.local`);
}
```

athielba (approved):
LGTM, 👏🏼 , except few feedbacks

apps/slack/vitest.config.mts
proquig: As slack doesn't have a table name `organisations` but `teams` instead, I kept this script here to delete the data from this table before each test.

packages/test-utils/src/vitest/global-setup.ts
elbalexandre: This is dope !

packages/test-utils/src/vitest/setup-database.ts
```diff
@@ -0,0 +1,15 @@
...
+import { drizzle } from 'drizzle-orm/postgres-js';
+import { sql } from 'drizzle-orm';
+
+// Delete every entries in the database between each tests
+beforeEach(async () => {
+  if (!process.env.DATABASE_URL) {
+    throw new Error('DATABASE_URL environment variable is not set');
+  }
+
+  const pg = postgres(process.env.DATABASE_URL);
```
elbalexandre, line 12: So the setup scripts are always running in node environment, good to know

elbalexandre (approved):

athielba (approved):

PR #213, opened 2024-04-11
Elastic Integration

mbliakharskyi:
## Description

## Checklist:


apps/elastic/src/common/apiclient.ts
```diff
@@ -0,0 +1,35 @@
+import { ElasticError } from '@/connectors/commons/error';
+
+export const getElasticApiClient = () => {
```
elbalexandre, line 3: Please don't do that kind of factorisation. It act as a black box when reading the connector code.

We might give tools to write less code later, but we need to discuss the implementation first.

apps/elastic/src/connectors/users.ts
```diff
@@ -0,0 +1,117 @@
...
+    throw new ElasticError('Could not retrieve account id');
+  }
+  const { id: accountId } = accounts[0];
+
+  return {
+    accountId: accountId.toString(),
+  };
+};
+
+export const deleteUsers = async ({ userIds, accountId, apiKey }: DeleteUsersParams) => {
```
elbalexandre, line 103: Let's delete user one by one (even if the API allows multiple user deletion at once)
```suggestion
export const deleteUser = async ({ userId, accountId, apiKey }: DeleteUsersParams) => {
```

apps/elastic/src/connectors/users.ts
```diff
@@ -0,0 +1,117 @@
...
+    method: 'DELETE',
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: `ApiKey ${apiKey}`,
+    },
+  });
+
+  if (!response.ok && response.status !== 400) {
+    throw new ElasticError(`Could not delete user with Id: ${userIds}`, { response });
+  }
```
elbalexandre, line 116:
```suggestion
const elasticErrorDataSchema = z.object({
  errors: z.array(z.object({
    code: z.string().nullish(),
    message: z.string().nullish()
  }))
})
// ...
  if (response.status === 400) {
    const errorDataResult = elasticErrorDataSchema.safeParse(await response.json())
    if (errorDataResult.success && errorDataResult.data.errors.at(0)?.code === 'organization.membership_not_found') {
      return;
    }
  }

  if (!response.ok) {
    throw new ElasticError(`Could not delete user with Id: ${userId}`, { response });
  }
```

apps/elastic/src/app/api/webhook/elba/users/delete/service.ts
```diff
@@ -0,0 +1,19 @@
+import { inngest } from '@/inngest/client';
+
+export const deleteUsers = async ({
```
elbalexandre, line 3: It should be `userId` here and in the inngest event (so in the function deleteUser too)

mbliakharskyi: okay.

apps/elastic/src/inngest/functions/users/delete-users.ts
```diff
@@ -0,0 +1,42 @@
...
+    const [organisation] = await db
+      .select({
+        apiKey: Organisation.apiKey,
+        accountId: Organisation.accountId,
+      })
+      .from(Organisation)
+      .where(eq(Organisation.id, organisationId));
+
+    if (!organisation) {
+      throw new NonRetriableError(`Could not retrieve ${userId}`);
```
elbalexandre, line 31:
```suggestion
      throw new NonRetriableError(`Could not retrieve organisation with id=${organisationId}`);
```

elbalexandre (approved):

PR #216, opened 2024-04-12
feat(box):  auth, sync users and delete user features

athielba:
## Description

Box complete setup
1. Authentication
2. Refresh  access token
3. Refresh access token before it expires
4. Sync user
5. Schedule sync users
6. Delete user

## Checklist:


apps/box/scripts/clean-database.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
elbalexandre, line 1: We probably don't need that for every integration.
I propose to keep integration simple now and we could add a shared script that do that later.

apps/box/src/app/auth/route.ts
```diff
@@ -0,0 +1,28 @@
...
+      region,
+      sourceId: env.ELBA_SOURCE_ID,
+      baseUrl: env.ELBA_REDIRECT_URL,
+      error: 'unauthorized',
+    });
+  }
+
+  await setupOrganisation({ organisationId, code, region });
+
+  redirect(env.ELBA_REDIRECT_URL, RedirectType.replace);
```
elbalexandre, line 27:
```suggestion
  redirect(
      getRedirectUrl({
        region,
        sourceId: env.ELBA_SOURCE_ID,
        baseUrl: env.ELBA_REDIRECT_URL,
      })
  , RedirectType.replace);
```

apps/box/src/app/auth/service.ts
```diff
@@ -0,0 +1,67 @@
...
+      accessToken: encodedAccessToken,
+      refreshToken: encodedRefreshToken,
+      region,
+    })
+    .onConflictDoUpdate({
+      target: organisationsTable.id,
+      set: {
+        accessToken: encodedAccessToken,
+        refreshToken: encodedRefreshToken,
+        region,
```
elbalexandre, line 37: My bad, I let that live in the template, but it's useless as both region have different DB.
```suggestion
```

apps/box/src/common/apiclient.ts
```diff
@@ -0,0 +1,45 @@
+import { BoxError } from '@/connectors/commons/error';
```
elbalexandre, line 1: As discussed, let's avoid this for now.

apps/box/src/database/schema.ts
```diff
@@ -0,0 +1,11 @@
+import { uuid, text, pgTable } from 'drizzle-orm/pg-core';
+import { type InferSelectModel } from 'drizzle-orm';
+
+export const organisationsTable = pgTable('organisations', {
+  id: uuid('id').primaryKey(),
+  region: text('region').notNull(),
+  accessToken: text('accessToken').notNull(),
+  refreshToken: text('refreshToken').notNull(),
+});
```
elbalexandre, line 9:
```suggestion
  createdAt: timestamp('created_at').default(sql`now()`).notNull(),
});
```

apps/box/src/inngest/functions/users/delete-users.ts
```diff
@@ -0,0 +1,39 @@
...
+import { decrypt } from '@/common/crypto';
+
+export const deleteSourceUsers = inngest.createFunction(
+  {
+    id: 'box-delete-users',
+    concurrency: {
+      key: 'event.data.organisationId',
+      limit: 1,
+    },
+    retries: 5,
```
elbalexandre, line 16: Let's put the 2 values in env var and bump concurrency number (we need to check the doc for RateLimit but we could go higher than that - 5 seems acceptable)

apps/box/src/inngest/functions/users/schedule-users-synchronize.ts
```diff
@@ -0,0 +1,36 @@
...
+export const scheduleUsersSynchronize = inngest.createFunction(
+  { id: 'box-schedule-users-syncs' },
+  { cron: env.USERS_SYNC_CRON },
+  async ({ step }) => {
+    const organisations = await db
+      .select({
+        id: organisationsTable.id,
+        region: organisationsTable.region,
+        accessToken: organisationsTable.accessToken,
+        refreshToken: organisationsTable.refreshToken,
```
elbalexandre, line 15:
```suggestion
```

apps/box/src/inngest/functions/users/synchronize-users.ts
```diff
@@ -0,0 +1,92 @@
...
+  {
+    id: 'box-synchronize-users',
+    priority: {
+      run: 'event.data.isFirstSync ? 600 : -600',
+    },
+    concurrency: {
+      key: 'event.data.organisationId',
+      limit: 1,
+    },
+    retries: 3,
```
elbalexandre, line 30: Same here

apps/box/src/inngest/middlewares/rate-limit-middleware.ts
```diff
@@ -0,0 +1,61 @@
...
+                retryAfter = parseInt(retryAfterHeader, 10);
+              }
+
+              return {
+                ...context,
+                result: {
+                  ...result,
+                  error: new RetryAfterError(
+                    `API rate limit reached by '${fn.name}', retry after ${retryAfter} seconds.`,
+                    retryAfter.toString(),
```
elbalexandre, line 48: We should put the time indicator, if we return a number it's parsed at millisecond, here it's seconds.
```suggestion
                    `${retryAfter.toString()}s`,
```

athielba: Good catch

athielba: I will use `ms` as others

apps/box/src/common/apiclient.ts
athielba: Maksym is using the same  type of Error class everywhere , I don't think we discussed about this  (We  discussed about  general rest client pattern., isn't it ?)

elbalexandre: I was talking about the apiclient not the error

apps/box/src/database/schema.ts
elbalexandre: It's convenient to debug issues in prod.

athielba: Are we going to apply this changes everywhere ?

proquig (approved):

PR #234, opened 2024-04-22
feat(doppler): auth & user sync feature

athielba:
## Description

1. Setup Doppler integration
2. Sync user
3. Schedule user sync

Screenshot 2024-04-30 at 10 27 33 AM
## Checklist:


elbalexandre (approved):

elbalexandre (commented):
The quality checks are not passing

apps/doppler/src/app/install/page.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  return (
+    <>
+      <h1>Setup Doppler integration</h1>
+      <InstructionsSteps>
+        <InstructionsStep index={1}>
+          <h3>How to Generate an API Key</h3>
+          <ol>
+            <li>
+              <strong>1. Log in to Your Account</strong>
```
elbalexandre, line 33: There is surely a way to make this tutorial shorter.

Perhaps we could skip the login part.

apps/doppler/src/app/install/page.tsx
```diff
@@ -0,0 +1,131 @@
...
+              <strong>7. Configure Workplace Role</strong>
+              <p>
+                Locate the <strong>Workplace Role</strong> section.
+              </p>
+              <ul>
+                <li>
+                  Change the role to <strong>Admin</strong> and then click <strong>Save</strong>.
+                </li>
+              </ul>
+            </li>
```
elbalexandre, line 84: I don't know what the SaaS looks like, but I guess step 6 is skippable.

apps/doppler/src/app/install/page.tsx
```diff
@@ -0,0 +1,131 @@
...
+              <strong>8. Set Project Access</strong>
+              <p>
+                Under <strong>Project Access</strong>, select the projects to which you want the
+                service account to have access.
+              </p>
+              <ul>
+                <li>
+                  9. Click <strong>Save</strong> to confirm the selections.
+                </li>
+              </ul>
```
elbalexandre, line 95: Do we really need to ask them to select project access to list workspace users ?

elbalexandre: Instead of making a list in the first step, why not adding more InstructionsStep like so:

(wording is trash and there is no <strong>)
```
Step 1 
<h3>create a service account</h3>
<p>On your Doppler account, go to tokens page.</p>
<p>In service tab click "Manage service accounts".</p>
<p>Click on "+" to create a new service account.</p>

Step 2 
<h3>generate an api token</h3>
<p>On your new service account, click on "+" to create a new token.</p>
<p>We encourage to leave the "Expire Token" checkbox unchecked as the integration will be disconnected after token expiration.</p>
<p>Copy the created token to the form bellow</p>

Step 3 - the form
```

athielba: It was an experimental version(ChatGpt) 😆 , I will update it

elbalexandre (approved):

apps/doppler/src/app/install/service.ts
```diff
@@ -0,0 +1,55 @@
...
+};
+
+export const registerOrganisation = async ({
+  organisationId,
+  apiToken,
+  region,
+}: SetupOrganisationParams) => {
+  const encodedApiToken = await encrypt(apiToken);
+
+  await getUsers({ apiToken });
```
proquig, line 20: Not a fan of this, especially when there is an endpoint to get token information https://docs.doppler.com/reference/auth-me
Sadly it doesn't seem to provide much information like when the token is expiring
Also we should validate
- the token comes from a service account
- the service account has access to the necessary permissions

proquig (approved):
All good, but IMO we should have stricter checks regarding tokens when tokens are manually provided on a form. We should:
- Make sure the token is working (tested in most case), by calling the specific token information when the source provides it (like here), and if it's not the case by calling a random endpoint like we do here (listing users)
- Check the token type, for example here from what I saw we can make sure the token is a service account api token
- Make sure the scopes/permissions are the one we expect
- Make sure the token doesn't have an expiration date
Ofc implement these checks when possible

PR #308, opened 2024-06-05
refactor(confluence): add new token refreshment strategy

elbalexandre:
## Description

- add `app-core` package
  - expose utils to handle common behaviour
- implement `app-core` refresh token strategy in confluence

## Additional Notes

I can remove the `app-core` package and implement this logic in confluence integration as well if the current implementation is not desired.

I will write test for `app-core` after getting an opinion on it (I have to mount a db to do some of the unit-test).

I will use this branch to test the new refresh token strategy on staging.

## Checklist:


apps/confluence/src/connectors/confluence/auth.test.ts
```diff
@@ -16,6 +16,18 @@ const oauthToken = {
...
+
+const organisation = {
+  id: `45a76301-f1dd-4a77-b12f-9d7d3fca3c90`,
+  region: 'us',
+  instanceId: '1234',
+  instanceUrl: 'http://foo.bar',
+  accessToken: 'encrypted-access-token',
+  refreshToken: 'encrypted-refresh-token',
+  createdAt: new Date(),
+  tokenExpiresAt: new Date(),
```
elbalexandre, line 29:
```suggestion
```

apps/confluence/src/connectors/confluence/auth.test.ts
```diff
@@ -85,6 +97,14 @@ describe('auth connector', () => {
...
 
   describe('getRefreshedToken', () => {
+    beforeAll(() => {
+      vi.setSystemTime(now);
+    });
+
+    afterAll(() => {
+      vi.useRealTimers();
+    });
+
```
elbalexandre, line 107:
```suggestion
```

packages/app-core/src/inngest/functions/refresh-token.ts
```diff
@@ -0,0 +1,133 @@
...
+        });
+      },
+    },
+    {
+      event: eventName,
+    },
+    async ({ event, step }) => {
+      const { organisationId, expiresAt, refreshAt } = event.data as RefreshTokenEventData;
+
+      if (refreshAt) {
```
elbalexandre, line 81: This params is present when the onFailure re-trigger the refresh

packages/app-core/src/inngest/functions/refresh-token.ts
```diff
@@ -0,0 +1,133 @@
...
+          throw new NonRetriableError(`Could not retrieve organisation with id=${organisationId}`);
+        }
+
+        const decryptedOrganisation = await decryptRecord({
+          data: organisation,
+          encryptedKeys,
+          encryptionKey,
+        });
+
+        const result = await getRefreshedToken(decryptedOrganisation);
```
elbalexandre, line 107: We could catch error thrown here to call a callback to know if the org should be uninstalled.

packages/app-core/src/inngest/functions/refresh-token.ts
```diff
@@ -0,0 +1,133 @@
...
+          'wait-before-expiration',
+          subSeconds(new Date(expiresAt), refreshAdvence)
+        );
+      }
+
+      const nextExpiresAt = await step.run('refresh-token', async () => {
+        const [organisation] = await db
+          .select()
+          .from(organisationsTable)
+          // @ts-expect-error -- too hard to infer
```
elbalexandre, line 94: It's doable, but IMO it's an overhead for now.

packages/app-core/src/inngest/functions/refresh-token.ts
```diff
@@ -0,0 +1,133 @@
...
+      ],
+      retries: 5,
+      onFailure: async ({ event, step }) => {
+        const { organisationId, expiresAt } = event.data.event.data as RefreshTokenEventData;
+        await step.sendEvent('reschedule-token-refresh', {
+          name: eventName,
+          data: {
+            organisationId,
+            expiresAt,
+            refreshAt: addSeconds(new Date(), failureBackoff).getTime(),
```
elbalexandre, line 70: Havin both params is convenient to debug on inngest dashboard.

packages/app-core/src/encryption.ts
```diff
@@ -0,0 +1,63 @@
+import { decryptText, encryptText } from '@elba-security/utils';
+
+export type EncryptedKeys<T> = (keyof {
+  [K in keyof T as T[K] extends string ? K : never]: K;
+})[];
+
```
elbalexandre, line 6: I had to use a newer version of typescript to make this magic tricks works.
Basically, it create an array of key of an object only on property that are string.
It's convenient to control which key can be encrypted. The autocomplete propose only theses one.

It's totally skipable, was just for DX.

PR #336, opened 2024-06-20
feat(zod): add zod package

elbalexandre:
## Description

- add zod package
  - `zInngestRetry`: schema that match inngest retry type
  - `zSegregate`:  schema builder that split array into `valids` & `invalids` entries
    - return an error only if the value is not array or valids / invalids length does not match given config
  - `zFilteredArray`: schema build that filter out array entries
    - return an error only if the value is not array or result array length does not match given config
 - implement zod package utils in `microsoft`

## Additional Notes

I am open to API changes + renaming (especially zSegregate that could be not explicit enough or not inclusive).
I would loved to have the ability to do stuff like `zFilteredArray().min(0)` but TS is a jerk and I could not make it work with zod.

## Checklist:

### DIFF
--- a/apps/microsoft/src/connectors/elba/third-party-apps/objects.ts
+++ b/apps/microsoft/src/connectors/elba/third-party-apps/objects.ts
@@ -1,16 +1,6 @@
 import type { ThirdPartyAppsObjectUser } from '@elba-security/sdk';
-import type {
-  MicrosoftAppPermission,
-  MicrosoftAppWithOauthGrants,
-} from '@/connectors/microsoft/apps';
+import type { MicrosoftAppWithOauthGrants } from '@/connectors/microsoft/apps';
 import type { AppUserMetadata } from './metadata';
-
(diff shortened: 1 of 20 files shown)


apps/microsoft/src/connectors/elba/third-party-apps/objects.ts
```diff
@@ -1,17 +1,7 @@
...
-import type {
-  MicrosoftAppPermission,
-  MicrosoftAppWithOauthGrants,
-} from '@/connectors/microsoft/apps';
+import type { MicrosoftAppWithOauthGrants } from '@/connectors/microsoft/apps';
 import type { AppUserMetadata } from './metadata';
 
-type ValidMicrosoftAppRolePermission = { id: string; principalId: string };
-
-const isValidAppRolePermission = (
```
elbalexandre, line 10: I do not need that anymore as I have used `z.filteredArray`.

PR #343, opened 2024-06-26
feat(sharepoint): sync and delete item permissions update

tGeorgiy:
## Description

Sync Item logic update, now permissions combined to one user-permissions instance, also delete item permissions logic updated.

## Checklist:

### DIFF
--- a/apps/sharepoint/src/connectors/microsoft/sharepoint/permissions.test.ts
+++ b/apps/sharepoint/src/connectors/microsoft/sharepoint/permissions.test.ts
@@ -1,4 +1,5 @@
 import { http } from 'msw';
+import { z } from 'zod';
 import { describe, expect, test, beforeEach, vi } from 'vitest';
 import { server } from '@elba-security/test-utils';
 import { env } from '@/common/env';
@@ -16,6 +17,14 @@ const siteId = 'some-site-id';
 const driveId = 'some-drive-id';
 const itemId = 'some-item-id';
 const permissionId = 'permission-id';
+
+const granteeSchema = z.object({
+  email: z.string().email(),
+});
+
+const dataSchema = z.object({
+  grantees: z.array(granteeSchema).min(1),
+});
 
 const permissions: MicrosoftDriveItemPermission[] = Array.from({ length: 5 }, (_, i) => ({
   id: `permission-id-${i}`,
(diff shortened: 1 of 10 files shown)


apps/sharepoint/src/connectors/microsoft/sharepoint/permissions.test.ts
```diff
@@ -17,6 +18,14 @@ const driveId = 'some-drive-id';
 const itemId = 'some-item-id';
 const permissionId = 'permission-id';
 
+const granteeSchema = z.object({
+  email: z.string().email(),
```
elbalexandre, line 22: I am not sure email with `some_guest#external@foo.bar` are passing the `z.string().email()` filter

elbalexandre: Looks like the schema used in permission is correct (z.string()).

apps/sharepoint/src/inngest/functions/data-protection/common/helpers.ts
```diff
@@ -56,39 +62,75 @@ export const getChunkedArray = <T>(array: T[], batchSize: number): T[][] => {
...
-      metadata: {
-        sharedLinks: [permission.link.webUrl],
-      },
-    };
-  }
+export const combinePermisisons = (
+  itemId: string,
+  permissions: MicrosoftDriveItemPermission[]
+): DataProtectionObjectPermission[] => {
+  const combinedArray: CombinedPermission[] = [];
```
elbalexandre, line 69:
```suggestion
  const combinedPermissions = new Map<string, CombinedPermission>();
```

Using a map will make function complexity `O(n)` instead of `O(n^2)`.

apps/sharepoint/src/inngest/functions/data-protection/common/helpers.ts
```diff
@@ -164,11 +206,9 @@ export const formatDataProtectionItems = ({
             driveId,
           } satisfies ItemMetadata,
           updatedAt: item.lastModifiedDateTime,
-          permissions: validPermissions
-            .map(formatPermissions)
-            .filter((permission): permission is DataProtectionObjectPermission =>
-              Boolean(permission)
-            ),
+          permissions: combinePermisisons(item.id, validPermissions).filter(
+            (permission): permission is DataProtectionObjectPermission => Boolean(permission)
```
elbalexandre, line 210: Why do we need to filter here ? Is there empty permission in the array ?

apps/sharepoint/src/inngest/functions/data-protection/common/helpers.ts
```diff
@@ -247,3 +287,102 @@ export const removeInheritedUpdate = (
...
+
+    if (permission.type === 'anyone') {
+      permissionDeletionArray.push({
+        permissionId: permission.id,
+      });
+    }
+
+    if (permission.type === 'user' && permission.metadata.linksPermissionIds?.length) {
+      permission.metadata.linksPermissionIds.forEach((permissionId) => {
+        const combinedLinkPermission = combinedLinkPermissions.find(
```
elbalexandre, line 358: Here complexity is O(n^3). We can do better.

apps/sharepoint/src/connectors/microsoft/sharepoint/permissions.test.ts
tGeorgiy: We can use string here also, but there should be only correct emails from microsoft, without special tags

apps/sharepoint/src/inngest/functions/data-protection/common/helpers.ts
tGeorgiy: I thought that the array there is not so large, and on small arrays there is almost no difference in speed, but yes, you can use the Map)

apps/sharepoint/src/inngest/functions/data-protection/common/helpers.ts
tGeorgiy: There shouldn't be any empty ones I think, I'll check, Guillaume added it before, I didn't pay attention.

apps/sharepoint/src/inngest/functions/data-protection/common/helpers.ts
tGeorgiy: I can use Map here to, it will be better

apps/sharepoint/src/inngest/functions/data-protection/common/helpers.ts
```diff
@@ -56,39 +62,71 @@ export const getChunkedArray = <T>(array: T[], batchSize: number): T[][] => {
...
+          metadata: {
+            directPermissionId: permission.id,
+            email: permission.grantedToV2.user.email,
+          },
+        });
+      } else {
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- will be there
+        const combinedItem = combinedPermissions.get(elbaPermissionId)!;
+        if (combinedItem.type === 'user') {
+          combinedItem.metadata.directPermissionId = permission.id;
```
elbalexandre, line 89: So we are overwritting the previous `directPermissionId`, why ?

tGeorgiy: No, it's not right, we have only one direct permission per item, so if we have permission before we will have linksPermissionIds and directPermissionId will be undefined. So here we are setting the directPermissionId.
It's not obvious, I can add a comment in the code about it.

elbalexandre (approved):

PR #344, opened 2024-06-26
chore(microsoft): add connectors logs

elbalexandre:
## Description

- add logs to connectors failed request's response

## Additional Notes

N/A

## Checklist:

### DIFF
--- a/apps/microsoft/src/connectors/microsoft/apps.ts
+++ b/apps/microsoft/src/connectors/microsoft/apps.ts
@@ -1,4 +1,5 @@
 import { z } from 'zod';
+import { logger } from '@elba-security/logger';
 import { env } from '@/env';
 import { MicrosoftError } from './commons/error';
 import type { MicrosoftPaginatedResponse } from './commons/pagination';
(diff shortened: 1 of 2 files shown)


apps/microsoft/src/connectors/microsoft/apps.ts
```diff
@@ -127,6 +128,12 @@ export const getApps = async ({ tenantId, token, skipToken }: GetAppsParams) =>
   });
 
   if (!response.ok) {
+    logger.error('Could not retrieve apps', {
+      response: {
+        status: response.status,
+        body: await response.text(),
```
proquig, line 134: Not sure you're using the body in middleware or somewhere else but you should clone it

elbalexandre: I am not using it, I guess we should clone in places that catch the response from the error. As the connector produce the response it should be able to do whatever it want.
Perhaps I should clone directly in the MicrosoftError.

proquig: Problem is that if someone use clone after the body has already been read on the IntegrationError it will be already be too late

elbalexandre: We can clone a response after it has been unwrapped.

proquig (approved):

PR #366, opened 2024-07-09
feat(teams): handle multiple organisations within same tenant

yuriiyf:
## Description
As we can have multiple organisations within same tenant (Microsoft Org) and therefor we need to handle it. The main problem is that MS have restrictions regarding how many subscriptions could be created for same resources in same tenant. 
So we should not create subscriptions for same resources in same tenant for different orgs and reuse one subscription for all orgs.

## Checklist:

### DIFF
--- a/apps/teams/src/database/schema.ts
+++ b/apps/teams/src/database/schema.ts
@@ -1,4 +1,4 @@
-import { uuid, text, timestamp, pgTable } from 'drizzle-orm/pg-core';
+import { uuid, text, timestamp, pgTable, primaryKey } from 'drizzle-orm/pg-core';
 
 export const organisationsTable = pgTable('organisations', {
   id: uuid('id').primaryKey(),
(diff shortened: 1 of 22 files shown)


apps/teams/src/database/schema.ts
```diff
@@ -8,21 +8,26 @@ export const organisationsTable = pgTable('organisations', {
...
+      .notNull(),
+    membershipType: text('membership_type').notNull(),
+    displayName: text('display_name').notNull(),
+  },
+  (table) => {
+    return {
+      pk: primaryKey({ columns: [table.id, table.organisationId] }),
+    };
+  }
+);
```
yuriiyf, line 26: Replacing combined primary key (`organisationId: channelId`) with composite primary key (id, organisationId). It's breaking change, so either we need to create manual migration or remove everything from table if it could be done (no users are currently using Teams data protection)

apps/teams/src/database/schema.ts
```diff
@@ -8,21 +8,26 @@ export const organisationsTable = pgTable('organisations', {
...
 
 export const subscriptionsTable = pgTable('subscriptions', {
   id: text('id').primaryKey(),
-  organisationId: uuid('organisation_id')
-    .references(() => organisationsTable.id, { onDelete: 'cascade', onUpdate: 'restrict' })
-    .notNull(),
+  tenantId: text('tenant_id').notNull(),
```
yuriiyf, line 30: Replace `organisationId` with `tenantId` as all our subscriptions created for tenant and therefor different organisations can have same subscriptions as they shared same tenant (MS ORG). It's also a breaking change, so either we use manual migration or remove everything from database.

PR #385, opened 2024-07-25
chore: include isSuspendable flag to sync user object

athielba:
## Description

`isSuspendable ` flag added to all the integrations that support user deletion

## Checklist:

### DIFF
--- a/apps/box/src/inngest/functions/users/synchronize-users.ts
+++ b/apps/box/src/inngest/functions/users/synchronize-users.ts
@@ -16,6 +16,7 @@ const formatElbaUser = (user: BoxUser): User => ({
   displayName: user.name,
   email: user.login,
   additionalEmails: [],
+  isSuspendable: true,
 });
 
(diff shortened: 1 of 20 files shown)


apps/box/src/inngest/functions/users/synchronize-users.ts
```diff
@@ -16,6 +16,7 @@ const formatElbaUser = (user: BoxUser): User => ({
   displayName: user.name,
   email: user.login,
   additionalEmails: [],
+  isSuspendable: true,
```
elbalexandre, line 19: Do we need to update test here ?

athielba: This is a default value for time being, Maksym will update it properly

proquig (commented):
Looks good to me but it seems that Jira is missing

proquig (approved):

PR #396, opened 2024-08-08
feat(asana): add isSuspendable param

mbliakharskyi:
## Description

Asana Integration
1. Added `isSuspendable` param
2. `url` is not possible because users are managed separately according to the workspaces.
 - A user is possible to belong to multiple workspaces.

## Additional Notes

N/A

## Checklist:

### DIFF
--- a/apps/asana/drizzle/0000_soft_winter_soldier.sql
+++ b/apps/asana/drizzle/0000_soft_winter_soldier.sql
@@ -3,5 +3,6 @@ CREATE TABLE IF NOT EXISTS "organisations" (
 	"region" text NOT NULL,
 	"created_at" timestamp DEFAULT now() NOT NULL,
 	"access_token" text NOT NULL,
+	"owner_id" text NOT NULL,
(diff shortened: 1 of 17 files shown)


apps/asana/drizzle/0000_soft_winter_soldier.sql
```diff
@@ -3,5 +3,6 @@ CREATE TABLE IF NOT EXISTS "organisations" (
 	"region" text NOT NULL,
 	"created_at" timestamp DEFAULT now() NOT NULL,
 	"access_token" text NOT NULL,
+	"owner_id" text NOT NULL,
 	"refresh_token" text NOT NULL
 );
```
athielba, line 8: @mbliakharskyi , since it is already merged with staging, you should not delete and regenerate the sql, please generate the migration file for the new field, do not modify the existing migrations file if it is already merged with staging

athielba (changes requested):

PR #401, opened 2024-08-12
feat(fivetran): add isSuspendable and url params

mbliakharskyi:
## Description

Fivetran Integration
- Added `isSuspendable` and `url` params
## Additional Notes

N/A

## Checklist:


apps/fivetran/drizzle/meta/0000_snapshot.json
```diff
@@ -1,5 +1,5 @@
 {
-  "id": "25f9abbf-62b8-4130-8dab-a886cd9b419d",
+  "id": "60c711ca-b45e-4eb2-bf54-445f149570a8",
```
elbalexandre, line 2: Fivetran is already deployed, in this case we need a new migration file

apps/fivetran/src/app/install/service.test.ts
```diff
@@ -3,39 +3,25 @@ import { eq } from 'drizzle-orm';
 import { db } from '@/database/client';
 import { organisationsTable } from '@/database/schema';
 import { inngest } from '@/inngest/client';
-import * as userConnector from '@/connectors/users';
-import type { FivetranUser } from '@/connectors/users';
-import { FivetranError } from '@/connectors/commons/error';
+import * as userConnector from '@/connectors/fivetran/users';
+import { FivetranError } from '@/connectors/common/error';
```
elbalexandre, line 7: (optionnal) That's nice that you rename `commons` to `common`, but it would be even better to put the common folder in `connectors/fivetran` because it's only common to `fivetran` (not elba)

mbliakharskyi: Got it!

PR #404, opened 2024-08-14
feat(zoom): add isSuspendable and url params

mbliakharskyi:
## Description

Zoom Integration
1. Added `isSuspendable` and `url` params
2. Can't deactivate a user with a role of `owner` and `admin`.

## Additional Notes

N/A

## Checklist:

### DIFF
--- a/apps/zoom/src/inngest/functions/users/sync-users.ts
+++ b/apps/zoom/src/inngest/functions/users/sync-users.ts
@@ -10,12 +10,17 @@ import { decrypt } from '@/common/crypto';
 import { type ZoomUser } from '@/connectors/zoom/users';
 import { createElbaClient } from '@/connectors/elba/client';
 
+enum ZoomUserRole {
+  Owner = '0',
+  Admin = '1',
+}
+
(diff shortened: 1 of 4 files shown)


apps/zoom/src/inngest/functions/users/sync-users.ts
```diff
@@ -15,7 +15,8 @@ const formatElbaUser = (user: ZoomUser): User => ({
   displayName: user.display_name,
   email: user.email,
   additionalEmails: [],
-  isSuspendable: true,
+  isSuspendable: user.role_id !== '0' && user.role_id !== '1', // '0': owner, '1: admin'
```
llanicm, line 18: Can we change this to a mapped enum, it would remove the need for the comment and make the code itself more readable. Also will make reuse of these values through the integration easier and keep them maintained in a central place.

mbliakharskyi: did it

elbalexandre (approved):

PR #405, opened 2024-08-16
feat(dbx-edge-tpa): third party apps sync included

athielba:
## Description

1. Sync third party apps
2. Refresh app
3. Delete App

## Checklist:


apps/dropbox-edge/src/connectors/dropbox/apps.ts
```diff
@@ -0,0 +1,126 @@
...
+  if (!response.ok) {
+    throw await DropboxError.fromResponse('Could not retrieve linked apps', { response });
+  }
+
+  const data: unknown = await response.json();
+
+  const result = linkedAppsSchema.omit({ team_member_id: true }).safeParse(data);
+
+  if (!result.success) {
+    throw await DropboxError.fromResponse('Could not parse linked apps response', { response });
```
elbalexandre, line 93: Not a fan of this, IMO the `result.error` is crucial to debug and understand which field is wrong / missing.
Happened to me while debugging confluence, having the zod errors logged is really convenient.

apps/dropbox-edge/src/connectors/dropbox/users.ts
```diff
@@ -105,7 +105,7 @@ const teamMemberSchema = z.object({
       display_name: z.string().min(1),
     }),
     status: z.object({
-      '.tag': z.union([z.literal('active'), z.literal('suspended')]),
+      '.tag': z.string(), // z.union([z.literal('active'), z.literal('suspended')]),
```
elbalexandre, line 108: Are sure about this ? Perhaps we can remove the comment if so.

apps/dropbox-edge/src/connectors/dropbox/users.ts
```diff
@@ -172,6 +172,11 @@ export const getUsers = async ({
     const userResult = teamMemberSchema.safeParse(member);
 
     if (userResult.success) {
+      // Only active users are valid
+      if (userResult.data.profile.status['.tag'] !== 'active') {
```
elbalexandre, line 176: [`zFilteredArray`](https://github.com/elba-security/elba-security/pull/336) could have been nice to use here :(

apps/dropbox-edge/src/connectors/dropbox/apps.ts
athielba: Dropbox return common error in the body, unless we log it properly, it is hard to debug, even I dont want to call another method to log the error but I don't find a better way do it

apps/dropbox-edge/src/connectors/dropbox/users.ts
athielba: Point, I added this while I was testing forgot to remove 🙈

apps/dropbox-edge/src/connectors/dropbox/users.ts
elbalexandre: Yes, it's just a remark

PR #410, opened 2024-08-20
feat(docusign): update the codebase and add isSuspendable

mbliakharskyi:
## Description

Docusign Integration
1. Updated the codebase
2. The owner can be deleted, so isSuspendable is True in all cases.
3. Tested sync users, pagination, and delete a user

## Additional Notes

N/A

## Checklist:


apps/docusign/src/inngest/client.ts
```diff
@@ -32,7 +32,7 @@ export const inngest = new Inngest({
     'docusign/users.delete.requested': {
```
llanicm, line 32:
```suggestion
    'docusign/user.delete.requested': {
```

apps/docusign/src/inngest/functions/users/delete-user.ts
```diff
@@ -3,14 +3,22 @@ import { NonRetriableError } from 'inngest';
...
-import { deleteUsers as deleteDocusignUsers } from '@/connectors/docusign/users';
+import { deleteUser as deleteDocusignUser } from '@/connectors/docusign/users';
 import { decrypt } from '@/common/crypto';
+import { env } from '@/common/env';
 
-export const deleteUsers = inngest.createFunction(
-  { id: 'delete-users' },
+export const deleteUser = inngest.createFunction(
+  {
+    id: 'docusign-delete-users',
```
llanicm, line 12:
```suggestion
    id: 'docusign-delete-user',
```

apps/docusign/src/inngest/functions/users/delete-user.ts
```diff
@@ -3,14 +3,22 @@ import { NonRetriableError } from 'inngest';
...
+    concurrency: {
+      key: 'event.data.organisationId',
+      limit: env.DOCUSIGN_DELETE_USER_CONCURRENCY,
+    },
+    retries: 5,
+  },
   { event: 'docusign/users.delete.requested' },
```
llanicm, line 19:
```suggestion
  { event: 'docusign/user.delete.requested' },
```

apps/docusign/src/connectors/docusign/users.ts
```diff
@@ -76,9 +76,7 @@ type DeleteUsersParams = {
   apiBaseUri: string;
```
llanicm, line 76: Change to `DeleteUserParams`

apps/docusign/src/connectors/docusign/users.ts
```diff
@@ -121,11 +119,11 @@ export const getUser = async ({
...
 };
 
-export const deleteUsers = async ({
+export const deleteUser = async ({
   apiBaseUri,
   accessToken,
   accountId,
-  users,
+  userId,
 }: DeleteUsersParams) => {
```
llanicm, line 127:
```suggestion
}: DeleteUserParams) => {
```

apps/docusign/src/connectors/docusign/users.ts
```diff
@@ -22,7 +22,7 @@ export type GetUsersParams = {
   accessToken: string;
   accountId: string;
   apiBaseUri: string;
-  page: string | null;
+  page?: string | null;
```
llanicm, line 25: What was the need for this change?

llanicm (commented):
Just some small changes.

apps/docusign/src/inngest/client.ts
mbliakharskyi: I think we need to keep the same code style. In other integrations, for example, `Calendly`, we used it.
`calendly/users.delete.requested`.

apps/docusign/src/connectors/docusign/users.ts
mbliakharskyi: I don't think it has to be that way.
When writing the test scripts, there's no need to pass `page: null` as a parameter.

apps/docusign/src/inngest/client.ts
```diff
@@ -32,7 +32,7 @@ export const inngest = new Inngest({
...
       data: {
         organisationId: string;
-        userIds: string[];
+        userId: string;
       };
```
athielba, line 36: Please don't change the logic, user deletion is workin fine in the previous code, you don't need to update any code relating tot user deleting

apps/docusign/src/inngest/client.ts
llanicm: Are the deleting single users or multiple users? I was more highlighting that the name currently suggests it can handle multiple when in fact is is only deleting a singular record at a time with the current implementation.

athielba: @mbliakharskyi  in  order to keep the consistency you can keep the same event name for everywhere despite of  multi user delete, we will update the `calendly`  later

apps/docusign/src/inngest/client.ts
mbliakharskyi: ok

PR #465, opened 2024-10-17
feat(database): Implementing the new database connector for Calendly

llanicm:
## Description

This show an example implementation of the new database connector in the Calendly integration.

## Checklist:


apps/calendly/src/inngest/functions/organisations/remove-organisation.ts
```diff
@@ -23,10 +22,10 @@ export const removeOrganisation = inngest.createFunction(
     const { organisationId } = event.data;
     const [organisation] = await db
       .select({
-        region: organisationsTable.region,
+        region: tables.organisationsTable.region,
```
athielba, line 25: I am wondering, Why do we need this extra change in this file, why can't we still use the `organisationsTable` directly from schema?

PR #466, opened 2024-10-23
feat(azuredevops): OAuth, sync users and user deletion.

mbliakharskyi:
## Description

Azure DevOps Integration.
Context: 
- It provides multiple organizations. So we need to redirect a user to select an organization he wants.
- To use an OAuth app, we need to set the organization-level policy setting. So we need to redirect the user to the organization settings page to set the option if it is OFF.

Implementation:
1. OAuth flow
2. Sync users
3. Delete a user

## Additional Notes

I implemented pagination feature based on the docs, but didn't test it because the default is 100.

## Checklist:


apps/azuredevops/src/connectors/azuredevops/users.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  const response = await fetch(url.toString(), {
+    method: 'GET',
+    headers: {
+      Authorization: `Bearer ${accessToken}`,
+      Accept: 'application/json',
+    },
+  });
+
+  if (!response.ok) {
```
athielba, line 143: @mbliakharskyi  You can simplify it like below
```
return {
     isInvalidSecuritySetting: !response.ok && response.status === 401
}
```

PR #471, opened 2024-11-05
feat(docusign): use nango

llanicm:
## Description

Context : https://elbateamworkspace.slack.com/archives/C03CJABEPF1/p1729239813015409
implementing the use of nango for authentication in the Docusign integration.

## Checklist:


llanicm:
Note: The landing page need some love :D

apps/docusign/src/inngest/functions/users/delete-users.ts
```diff
@@ -33,13 +32,26 @@ export const deleteUsers = inngest.createFunction(
...
-      users: userIds.map((userId) => ({ userId })),
-      accountId: organisation.accountId,
-    });
+      if (!credentials) {
+        throw new NonRetriableError(`Could not retrieve credentials`);
+      }
+
+      const { baseUri, accountId } = await getRequestInfo(
+        credentials.access_token,
+        env.DOCUSIGN_ROOT_URL
```
athielba, line 44: the root url is not constant for DocuSign, the development and production urls are different, that is the reason I decided to fetch from api and store in the db

apps/docusign/src/inngest/functions/users/sync-users.ts
```diff
@@ -80,19 +79,27 @@ export const syncUsers = inngest.createFunction(
...
+        accessToken: credentials.access_token,
         accountId,
-        apiBaseUri,
+        apiBaseUri: baseUri,
         page,
       });
 
       const users = result.validUsers.map((user) => {
-        return formatElbaUser({ user, authUserId: organisation.authUserId });
+        return formatElbaUser({ user, isSuspendable: user.userId !== ownerId });
```
athielba, line 102: I would keep the condition where it was, because it will give clear view what is happening for the object

packages/nango/.eslintrc.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": ["@elba-security/custom/library"],
+  "rules": {
+    "@typescript-eslint/no-unsafe-call": "off",
```
athielba, line 4: What is the reason to off the rules ?

packages/nango/src/auth/index.ts
```diff
@@ -0,0 +1,25 @@
+import { EventEmitter } from '@elba-security/utils';
```
athielba, line 1: I would use , `async/await`

```javascript
import { EventEmitter } from '@elba-security/utils';
import { createClient } from '../clients/frontend';

export const authenticate = async () => {
  if (!process.env.NANGO_INTEGRATION_ID) {
    throw new Error('NANGO_INTEGRATION_ID is not set');
  }

  const emitter = new EventEmitter();
  const nango = createClient();

  try {
    const result = await nango.auth(process.env.NANGO_INTEGRATION_ID, {
      detectClosedAuthWindow: true,
    });
    await emitter.emit('success', { ...result });
  } catch (error) {
    await emitter.emit('error', error);
  }

  return emitter;
};
```

packages/nango/src/clients/frontend.ts
```diff
@@ -0,0 +1,13 @@
+import Nango from '@nangohq/frontend';
+
+export const createClient = () => {
+  const publicKey = process.env.NANGO_PUBLIC_KEY;
+
+  // if (!publicKey) {
```
athielba, line 6: should be removed

packages/nango/src/credentials/index.ts
```diff
@@ -0,0 +1,34 @@
...
+    `https://${process.env.NANGO_HOST}/connection/${connectionId}?provider_config_key=${process.env.NANGO_INTEGRATION_ID}`,
+    {
+      headers: {
+        Authorization: `Bearer ${process.env.NANGO_SECRET_KEY}`,
+      },
+    }
+  );
+
+  if (!result.ok) {
+    return {
```
athielba, line 22: is it possible to get the complete error message?,  for example 403 may have different cases & it would be nice to give clear error message to know the exact issue

packages/nango/src/auth/index.ts
proquig: Yeah it's not needed, we can even get rid of the EventEmitter

proquig (approved):

PR #473, opened 2024-11-06
fix(google): improve error detection

proquig:
## Description

This improves Google integration error detection.
During the installation process, we now detect some known errors and redirect the user to a minimal error page displaying information about the error.
Currently there are 2 known errors:
- When the users is not a super admin of the Google Workspace, which was previously supported
- When the Google Drive access is restricted through the Google Workspace settings, which was not properly handled before.

Instead of redirecting to elba directly, the user is now redirected to this error page which should make it much more explicit about the kind of issue encountered as well as some tips to help the user fix it.

In addition to detect these errors early during the installation flow, some checks were also added early during the sync process:
- When syncing users, there is an early check to make sure the user is still admin
- When syncing drives, there is an early check to make sure the access to Google Drive is not restricted

These checks will remove the organisation and set an elba source connection error if they fail.

Other changes:
- Use design system component package
- Use tailwind css

## Checklist:


apps/google/src/inngest/functions/users/sync.ts
```diff
@@ -53,6 +53,9 @@ export const syncUsers = inngest.createFunction(
 
     const { users, nextPageToken: nextPage } = await step.run('list-users', async () => {
       const authClient = await getGoogleServiceAccountClient(googleAdminEmail, true);
+
+      await checkUserIsAdmin({ userId: googleAdminEmail, auth: authClient });
```
athielba, line 57: I am wondering, why do you want to check the admin privilege every time?  If the admin is removed or if the admin privilege is removed then we should completely remove the organisation , isn't it ? because this is not only the case for Dpd but also for everywhere

proquig: I'm not sure to understand your comment but the purpose of this check is to completely remove the organisation if the user is not admin anymore. I did the check here as we need admin privileges to sync users and the sync is triggered every weekday (more often than dp for example)

athielba (approved):

PR #480, opened 2024-11-15
feat(data-protection): limit trial organisation issues count [E-2384]

yuriiyf:
## Description
[E-2384](https://linear.app/elba/issue/E-2384/have-a-limited-number-of-issues-for-trial-organisations)

## Additional Notes
There basically two options of handling data-protection errors in current integrations:
1) Use inngest middleware that will listen specific errors (from Elba api) and throw `NonRetriableError` to abort current function, which will also trigger all `onFailure` listeners, so the scan will be finished fully without stayed function (when using `waitFor`). **But this will mean that almost all functions will be called**
2) Use util function to handle error, but we need a way to fully stop/finish sync, which means we will need to add extra logic to send an event that will identify that sync of current channel finished. In that case we can send an event to parent waiter `waitFor` to finish sync. **This way fewer functions are called, but it requires adding logic to each integration.**

## Checklist:

### DIFF
--- a/packages/inngest/src/middlewares/api.ts
+++ b/packages/inngest/src/middlewares/api.ts
@@ -1,0 +1,55 @@
+import { InngestMiddleware } from 'inngest';
+import {
+  ElbaError,
+  DataProtectionErrorCode,
+  type DataProtectionUpdateFailure,
+} from '@elba-security/sdk';
+
+const hasOrganisationIdProperty = (data: unknown): data is { organisationId: string } =>
+  typeof data === 'object' &&
+  data !== null &&
+  'organisationId' in data &&
+  typeof data.organisationId === 'string';
+
+export const createDataProtectionApiMiddleware = (cancelEventName: string) =>
+  new InngestMiddleware({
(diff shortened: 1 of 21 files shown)


packages/inngest/src/middlewares/api.ts
```diff
@@ -0,0 +1,51 @@
...
+                result: { error, ...result },
+                ...context
+              } = ctx;
+              const isElbaApiError = error instanceof ElbaError;
+
+              if (!isElbaApiError || !error.response) {
+                return ctx;
+              }
+
+              const response = (await error.response.json()) as DataProtectionUpdateFailure;
```
elbalexandre, line 30:
```suggestion
              const response = (await error.response.clone().json()) as DataProtectionUpdateFailure;
```

proquig (changes requested):
What we should do is create a new event `[integration]/sync.cancel`similar to `[integration]/app.uninstalled`. Every functions related to data protection sync should have a cancelOn event matching the organisationId

PR #621, opened 2025-04-07
chore(google): personal drive sync optimization

proquig:
## Description

This improves Google data protection sync. We now use the `shared` file attribute to know if a file is shared or not. This attribute only works for personal drives. When the file is not shared, we avoid retrieving its permissions and sending it. By consequence, personal drives sync speed should be improved

## Checklist:

Before you create this PR, confirm all the requirements listed below by
checking the checkboxes like this (`[x]`).

- [x] I have commented my code, particularly in hard-to-understand
areas.


athielba (approved):
LGTM, Except one question.
Don't we need to update  the test ?

